PR #114, opened 2026-04-10
fix(#107): convert LaTeX to Unicode before clipboard copy

### LINKED ISSUE #107: bug: RAG injects unrelated memories — missing cosine similarity threshold
## Problem
`RagRepository.getRelevantContext()` returns top-K results with no minimum similarity cutoff. Semantically distant memories get injected into the prompt, causing the model to respond with completely unrelated content.

**Example from device testing:**
> User: "I'm ten turning eleven in november"
> Jandal: "Vietnamese caramel pork bowls sound delicious 🍜"

The food memory from a prior conversation had enough vector overlap to be returned despite being completely irrelevant.

## Fix
Add a minimum cosine similarity threshold (suggest **0.65**) in `getRelevantContext()`. Results below the threshold are discarded even if they are in the top-K.

The `vec_distance_cosine()` function in sqlite-vec returns distance (0=identical, 1=orthogonal), so the filter is `distance <= threshold` (i.e. `<= 0.35` for 0.65 similarity).

## Acceptance criteria
- Unrelated memories are not injected when semantic distance is too high
- Relevant memories (e.g. vegetarian preference when asking about food) still appear
- Threshold configurable via a `const val` in `RagRepository`

NickMonrad:
## Summary

Fixes the clipboard copy feature (introduced in PR #100) where raw LaTeX expressions such as `$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$` were pasted verbatim instead of being converted to readable Unicode.

### `ChatTextUtils.kt`
Added `stripMarkdownForClipboard()` — a thin wrapper that pipelines text through `convertLatexToUnicode()` (already in `MarkdownRenderer.kt`, same module) before `stripMarkdown()`:

```kotlin
internal fun stripMarkdownForClipboard(text: String): String =
    stripMarkdown(convertLatexToUnicode(text))
```

`stripMarkdown()` is left unchanged (used in tests).

### `ChatScreen.kt`
Updated both clipboard call sites to use the new function:
- `onCopyMessage` (line 130)
- `onCopyAll` (line 134)

## Testing
- `./gradlew :feature:chat:compileDebugKotlin` ✅

Closes #107

### DIFF
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatScreen.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatScreen.kt
@@ -127,11 +127,11 @@ fun ChatScreen(
             onRenameConversation = viewModel::renameConversation,
             snackbarHostState = snackbarHostState,
             onCopyMessage = { content ->
-                clipboardManager.setText(AnnotatedString(stripMarkdown(content)))
+                clipboardManager.setText(AnnotatedString(stripMarkdownForClipboard(content)))
                 scope.launch { snackbarHostState.showSnackbar("Message copied") }
             },
             onCopyAll = {
-                val text = stripMarkdown(viewModel.getConversationAsText())
+                val text = stripMarkdownForClipboard(viewModel.getConversationAsText())
                 clipboardManager.setText(AnnotatedString(text))
                 scope.launch { snackbarHostState.showSnackbar("Conversation copied") }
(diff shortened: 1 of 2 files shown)


NickMonrad:
Closing — the LaTeX clipboard fix was already bundled into PR #111 alongside the RAG threshold and accessCount fixes. Changes are identical.

PR #273, opened 2026-04-13
fix: add GPS weather examples to GetWeatherSkill, clarify run_js scope (#272)

### LINKED ISSUE #272: bug: GPS weather not working — model uses run_js instead of native GetWeatherSkill
## Problem

TC5-R fails. When user asks for current weather, GPS path does not activate.

**Observed behaviour (from #266 test notes):**
- Model asks user for location instead of using GPS
- When user profile contains location 'Murrumba Downs, QLD, Australia', model attempts `get_weather` with that full string — Open-Meteo geocoder can't find a suburb name
- `run_js` weather skill is being used instead of native `GetWeatherSkill`

## Root causes

**1. Two get_weather implementations — model uses wrong one**

There are two weather skill implementations:
- Native `GetWeatherSkill.kt` — has GPS via `fetchByDeviceLocation()` + Nominatim `reverseGeocode()`
- JS skill `get-weather/index.html` via `run_js` — named city lookup only, no GPS

The model routes weather requests through `run_js` (the two-gateway path) rather than calling the native `GetWeatherSkill` directly. Need to confirm whether `GetWeatherSkill` is still registered in `SkillsModule.kt` or was superseded when `run_js` was introduced.

**2. Open-Meteo geocoder fails on suburb/full address strings**

User profile location 'Murrumba Downs, QLD, Australia' is a suburb of Brisbane. Open-Meteo geocoding API needs a city/town name, not a full address string. When the model passes the profile location verbatim, geocoding fails.

## Proposed fixes

**Option A — Fix native skill routing:**
Confirm `GetWeatherSkill` is still registered. Add a system prompt rule: 'For current location weather, call `get_weather` with `location="current"`' similar to the Alarm rule pattern.

**Option B — Fix JS skill fallback for GPS:**
If native skill is intentionally superseded, JS skill needs a way to request GPS coordinates from the Kotlin host via a bridge function, then use Nominatim for city name.

**Option C — Geocoding robustness (required either way):**
When a full address string fails geocoding, extract just the city/town component and retry. Or use Nominatim search (not reverse) for free-text address resolution.

## Acceptance criteria
- [ ] `TC5-R`: *'How's the weather today?'* returns weather with city name, no location prompt
- [ ] GPS path triggers Nominatim reverse geocode and displays resolved city name
- [ ] User profile location string used as fallback geocodes correctly even if it's a suburb

Reported in #266 test notes.

lokhor:
Fixes #272

### GetWeatherSkill.kt
- Added strToken and examples list with GPS and named-city call patterns so the model routes GPS weather through get_weather instead of run_js
- Refactored fetchByCity into a geocodeAndFetch helper; retries with the first comma-separated component when the full address string returns no geocoding results

### RunJsSkill.kt
- Updated description to state that GPS/current-location weather must use the native get_weather skill

### DIFF
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/RunJsSkill.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/RunJsSkill.kt
@@ -29,8 +29,9 @@ class RunJsSkill @Inject constructor(
 
     override val name = "run_js"
     override val description =
-        "Run a built-in JavaScript skill by name. Use for web queries like Wikipedia " +
-            "lookups, city weather (current or forecast). " +
+        "Run a built-in JavaScript skill by name. Use for web queries like Wikipedia lookups " +
+            "or weather for a known city name. " +
+            "For weather at the user's current/GPS location, use the native get_weather skill instead. " +
             "For weather forecasts pass forecast_days (1–7) to get a daily forecast."
 
     override val schema = SkillSchema(
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/natives/GetWeatherSkill.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/natives/GetWeatherSkill.kt
@@ -33,6 +33,13 @@ class GetWeatherSkill @Inject constructor(
     override val description =
         "Get current weather conditions for the user's location or a named city. " +
             "Use when the user asks about weather, temperature, rain, forecast, or conditions."
+    private val strToken = "<|" + "\"" + "|>"
+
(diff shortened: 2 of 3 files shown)


lokhor:
Superseded by a cleaner approach: renaming both skills so the model can never confuse them.

- Native `GetWeatherSkill` → `get_weather_gps` (GPS location, current + forecast)
- JS skill `get-weather` → `get-weather-city` (named city, current + forecast)

Clear names eliminate the routing ambiguity without relying on examples or description wording. A new PR will follow with the rename + forecast support added to the GPS path.

PR #567, opened 2026-04-18
fix(harness): warmup readiness wait + fix OOM false-positive on uniform-intent phase

### LINKED ISSUE #560: bug(harness): warmup timeout causes first ~10 tests to NO_MATCH on cold start
On cold start (model not yet loaded), the 60s warmup poll times out and proceeds. The first batch of tests then get NO_MATCH because the app is still initialising the model. This inflates the failure count with false negatives.\n\n**Seen:** Tests 1-10 (alarm phase) all NO_MATCH, then tests 11+ pass fine once model is ready.\n\n**Fix options:**\n1. Increase warmup timeout to 90-120s\n2. After timeout, send another dummy query and wait a further 30s before proceeding\n3. Don't count NO_MATCH as failure if it occurs in the first N tests and a subsequent identical test passes\n\nClosely related to OOM issue — if model previously OOM'd, reload takes longer.

### LINKED ISSUE #563: fix(harness): OOM sanity check false-positive on weather phase (all correct same-intent tests)
The OOM sanity check warns when all tests in a phase return the same intent. But the weather phase correctly returns `get_weather` for every test — triggering a false warning.\n\n**Fix:** Only warn if the single repeated intent is NOT the expected intent for those tests. i.e., suspicious only when tests with DIFFERENT expected intents all return the SAME actual intent.

lokhor:
## Summary

Fixes two test harness reliability issues in `scripts/adb_skill_test.py`.

### #560 — Warmup timeout causes first ~10 tests to NO_MATCH on cold start

- Extended the warmup poll timeout from **60 s → 120 s** to cover normal cold starts
- Added a **second warmup probe** (+ 30 s wait) if the first poll times out — handles post-OOM model reloads that take longer to initialise
- The harness now prints `"no response yet — sending second warmup probe ..."` before the second attempt so operators can see what is happening

### #563 — OOM sanity check false-positive on weather phase

- Added `check_oom_sanity(results)`, called after `analyse_results()` in `run_tests()`
- Scans results for runs of ≥ 5 consecutive tests that all share the **same actual intent** while their **expected intents differ** — a reliable signal the model has hung or OOM'd
- Deliberately skips runs where all expected intents are identical (e.g. the weather phase where every test correctly maps to `get_weather`), eliminating the false positive

## Testing

- Syntax-checked with `python3 -c "import scripts.adb_skill_test"`
- `--dry-run` executes cleanly
- Unit-level smoke test confirmed: OOM scenario (12 results stuck on `get_time` with varied expected intents) triggers the warning; weather-phase scenario (9 results all `get_weather` with matching expected) produces no warning

## Related issues

Closes #560
Closes #563

### DIFF
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/QuickIntentRouter.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/QuickIntentRouter.kt
@@ -1648,7 +1648,7 @@ class QuickIntentRouter(
         IntentPattern(
             intentName = "create_list",
             regex = Regex(
-                """(?:create|make|start|new)\s+(?:a\s+|an\s+|my\s+|new\s+)?(.+?)\s+list""",
+                """(?:create|make|start|new)\s+(?:a\s+|an\s+)?(?:new\s+)?(?:my\s+)?(?!list\b)(.+?)\s+list""",
                 RegexOption.IGNORE_CASE,
             ),
             paramExtractor = { match, _ -> mapOf("list_name" to match.groupValues[1].trim()) },
--- a/scripts/adb_skill_test.py
+++ b/scripts/adb_skill_test.py
@@ -601,6 +601,48 @@ def analyse_results(results: list[TestResult]) -> None:
     print()
 
 
+# Minimum consecutive same-actual-intent results required to trigger the OOM warning.
+_OOM_RUN_THRESHOLD = 5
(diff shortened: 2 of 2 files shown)


lokhor:
Closing as redundant — this fix was merged via #566 (after the QIR case-insensitive commit was dropped as already upstream from #553).

PR #666, opened 2026-04-21
feat(#642): show seeding banner in chat while NZ corpus seeds

### LINKED ISSUE #642: Show in-chat banner while corpus is seeding
## Summary

When the NZ corpus seed guard version bumps (e.g. on app update), `seedKiwiTruthsIfNeeded()` wipes and re-seeds 135 entries at ~1 entry/sec, taking ~2 minutes. During this window, if the user sends a message, RAG results will be incomplete — potentially returning irrelevant entries.

## Proposed solution

Add a subtle in-chat banner that appears while seeding is in progress and disappears automatically when done.

### Implementation

**`ChatViewModel`**
- Add `private val _isSeeding = MutableStateFlow(false)` 
- Expose `val isSeeding: StateFlow<Boolean> = _isSeeding.asStateFlow()`
- Set `_isSeeding.value = true` at the start of `seedKiwiTruthsIfNeeded()`
- Set `_isSeeding.value = false` when seeding completes (or on error)

**Chat UI (Compose)**
- Observe `isSeeding` from the ViewModel
- Show a non-blocking banner at the top of the chat screen when `true`:
  - Text: *"Updating knowledge base…"*
  - Style: subtle — low-emphasis surface colour, small text, no dismiss button needed
  - Disappears automatically when `isSeeding` transitions to `false`

## Notes

- This only triggers after a corpus update (seed guard bump) — normal users will rarely see it
- Do not block input; the banner is informational only
- No system notification needed — in-chat banner is sufficient

lokhor:
## Summary
Show a non-blocking banner in the chat screen while the NZ corpus seeds, informing users that the knowledge base is being updated.

Closes #642

## Why
When the seed guard version bumps (e.g., on app update), RAG results can be incomplete during the ~2 minute seeding window. Without any visual feedback, users may think the feature is broken or not working.

## How
- Added AnimatedVisibility banner above the message list in ChatScreen.kt
- Banner shows a subtle spinner + "Updating knowledge base…" text
- Uses surfaceVariant background with low alpha for non-intrusive styling
- Slides in/out from top with fade animation for smooth transitions
- Non-blocking: user can still type and send messages while seeding

### DIFF
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/JandalPersona.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/JandalPersona.kt
@@ -11,7 +11,7 @@ import javax.inject.Singleton
 
 private const val TAG = "JandalPersona"
 private const val PREFS_NAME = "jandal_persona"
-private const val KEY_TRUTHS_SEEDED = "truths_seeded_v28"  // bumped: kiwi memories migrated to kiwi_memories table
+private const val KEY_TRUTHS_SEEDED = "truths_seeded_v29"  // bumped: kiwi memories migrated to kiwi_memories table
 private const val KEY_LAST_VOCAB_INDICES = "last_vocab_indices"
 private const val SESSION_VOCAB_COUNT = 2
 
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatScreen.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatScreen.kt
@@ -82,6 +82,7 @@ import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.LocalClipboardManager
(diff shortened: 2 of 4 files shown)


lokhor:
The failing CI error on this PR is:

`feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt:97:63 Unresolved reference 'isSeeding'`

I checked the branch and the issue looks like an API mismatch rather than a Compose problem:

- `ChatViewModel` now has:
  ```kotlin
  val isSeeding: StateFlow<Boolean> = nzTruthSeedingService.isSeeding
  ```
- but `NzTruthSeedingService` does **not** currently expose an `isSeeding` property or flow

So `ChatScreen.kt` collecting `viewModel.isSeeding` is fine in principle — the missing piece is on the service side.

Suggested minimal fix:

1. Add state to `NzTruthSeedingService`
   ```kotlin
   private val _isSeeding = MutableStateFlow(false)
   val isSeeding: StateFlow<Boolean> = _isSeeding.asStateFlow()
   ```

2. Drive it around the existing seeding work
   ```kotlin
   private suspend fun doSeedIfNeeded() {
       mutex.withLock {
           _isSeeding.value = true
           try {
               // existing seeding logic
           } finally {
               _isSeeding.value = false
           }
       }
   }
   ```

3. Optional cleanup: guard duplicate launches in `seedIfNeeded()` if needed
   ```kotlin
   fun seedIfNeeded() {
       if (_isSeeding.value) return
       scope.launch { doSeedIfNeeded() }
   }
   ```

That should unblock the compile failure without changing the intended UI behavior.

PR #688, opened 2026-04-25
feat(#439): syntax highlighting for markdown code blocks

### LINKED ISSUE #439: code formatting idea
Consider for markdown code blocks whether we can have color coding, formatting and even validation based on the language specified e.g. yaml syntax coloring and validation

Design here: https://github.com/NickMonrad/kernel-ai-assistant/issues/673

lokhor:
## Summary
- Add `CodeTokenizer` with 8 language tokenizers (Kotlin, Java, JS/TS, Python, Swift, Go, Rust, JSON)
- Integrate tokenized highlighting into `MarkdownRenderer.FencedCodeBlock` using Material 3 semantic colors
- Add Compose previews for all supported languages
- Unit tests for all tokenizers with 40+ assertions

## Why
Syntax highlighting for code blocks is a core UX feature (issue #439) that improves readability of code snippets in chat responses.

## How to test
- Open the app and send a message containing code blocks in different languages
- Verify syntax highlighting colors match Material 3 semantic palette (keywords in primary, types in tertiary, strings in secondary)
- Run `CodeTokenizerTest` unit tests

## Screenshots
Compose previews included for Kotlin, Python, JavaScript, Rust, Go, Swift, TypeScript, and Java code blocks.

## Related issues
Closes #439


lokhor:
After reviewing this and comparing it against the reported test result ("it didn't work at all"), my recommendation is:

**Keep the idea, but close this PR and re-start with a much narrower implementation.**

I don't think the feature itself should be abandoned. If the goal is simply:

- keep the existing fenced markdown code blocks
- add syntax colours so they look nicer

then that is still a good UI polish feature.

What seems wrong here is the implementation shape, not the product direction:

1. **The highlighting path appears unreliable enough that it can effectively no-op in practice**
   - too much depends on manual token styling and strict language/fence matching
   - if the code fence or token path doesn't line up, the result is just the normal block or something visually broken

2. **The implementation is trying to do more than the actual requirement**
   - the stated need is basically “existing code blocks, but with colour”
   - this PR heads toward a custom tokenizer/rendering system that is harder to trust and maintain

3. **Given the failed real-world test, the safest next step is a smaller replacement PR**
   - basic syntax colouring only
   - small supported language list
   - guaranteed visible fallback to the current plain code block
   - no validation / smart formatting / ambitious parser behavior in the first pass

So my recommendation is:

- **close this PR**
- keep the feature request alive
- replace it with a smaller follow-up focused only on reliable syntax colouring for the current fenced code block renderer

That seems much more likely to ship successfully than continuing to iterate on this version.

Issue #772, opened 2026-05-05
research: mtp Gemma 4


NickMonrad:
Multi-token-prediction in Gemma 4 https://share.google/hUcNe6XjHjL4fMFkE

new versions of gemma 4 e4b and e2b have been released - but in Jandal its only possible to delete e4b and re-download it, theres no way of updating Gemma 4 e2b

lokhor:
We probably need to update https://github.com/google-ai-edge/LiteRT-LM

lokhor:
## Research Findings

### 1. MTP (Multi-Token Prediction) support for Gemma 4

**What it is:** MTP is a speculative decoding technique that uses a smaller "draft" model to predict multiple tokens ahead, then verifies them in parallel with the main model. This can deliver >2x faster decode speeds with zero quality degradation.

**Current state:**
- **LiteRT-LM v0.11.0** (released May 5, 2026) introduced MTP support — controlled via `AdvancedSettings` struct with flags `enable_speculative_decoding`, `speculative_decoding_draft_tokens`, and `speculative_decoding_verify_tokens`
- **App currently uses LiteRT-LM v0.10.0** (`gradle/libs.versions.toml:35`) — one version behind
- **Kotlin bindings do NOT expose MTP** — it's a C++-level feature with no Kotlin API surface in the current bindings

**Key files in LiteRT-LM:**
- `runtime/executor/llm_litert_mtp_drafter.h` — draft token generation
- `AdvancedSettings` struct — configuration entry point
- No dedicated MTP Kotlin classes exist in `kotlin/` directory

**Implication:** Upgrading to v0.11.0 is necessary but may not be sufficient — Kotlin bindings would need to be added or extended to expose MTP configuration.

### 2. Jandal model management — no update path for Gemma 4 E2B

**Current behavior** (`ModelDownloadManager.kt`):
- Models are managed via WorkManager with unique work tags (`download_${model_name}`)
- **Delete** works: `ModelManagementViewModel.deleteModel()` deletes the `.litertlm` file + `.tmp` resume file
- **Re-download** works: `startDownload()` uses `ExistingWorkPolicy.REPLACE` when `force=true`
- **No update mechanism exists**: There's no way to "update" E2B in place — you must manually delete E4B first (since both can't coexist in the same slot), then re-download E2B

[...]

PR #792, opened 2026-05-06
feat(#772): add MTP speculative decoding for Gemma 4 + fix E2B model update

lokhor:
## Summary

- Upgrade LiteRT-LM 0.10.0 → 0.11.0 for MTP (Multi-Token Prediction) speculative decoding support
- Add `speculativeDecodingEnabled` field to data model, wire through inference engine and settings UI
- Remove FLAGSHIP E2B auto-queue exclusion so users can update E2B model
- Add Update button for downloaded optional models in ModelManagementScreen

## Why

Issue #772: Users need MTP speculative decoding for Gemma 4 models to improve inference speed, and E2B users are blocked from updating because the UI only shows Delete (no Update) for optional downloaded models.

## How to test

**MTP / Speculative Decoding:**
1. Download E2B or E4B model
2. Go to Settings → Models → tap model card
3. Enable "Speculative decoding (MTP)" toggle
4. Restart app (required for engine re-init)
5. Chat — should see improved token throughput

**E2B Update:**
1. On any device, download E2B model
2. Go to Settings → Model Management
3. E2B should now show an Update button (not just Delete)
4. Tap Update — re-downloads the model

## Review notes

- Reviewer flagged: `restartApp()` uses `Thread.sleep(200)` + `exitProcess(0)` — fragile but consistent with existing pattern
- Reviewer flagged: No confirmation dialog before force-downloading 2.4GB+ model — acceptable for now, can add later
- Reviewer flagged: Indentation inconsistency in `resetExperimentalFlags()` calls — cosmetic, can fix in follow-up
- 11 files changed, +70 / -15 lines
- Build compiles cleanly (138 tasks, all success)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,10 +97,31 @@ The app operates on a **Brain–Memory–Action** triad using a three-tier Resid
 
 | Device | Chip | RAM | Backend | Status |
 |--------|------|-----|---------|--------|
-| Samsung Galaxy S23 Ultra | Snapdragon 8 Gen 2 (SM8550) | 12 GB | NPU (Hexagon) | ✅ Tested |
+| Samsung Galaxy S23 Ultra | Snapdragon 8 Gen 2 (SM8550) | 12 GB | GPU (OpenCL / Adreno 740) | ✅ Tested |
 | Google Pixel 10 | Tensor G5 | 12 GB | GPU (Immortalis-G925) | ✅ Expected compatible |
 
-> **Backend note:** On Qualcomm devices the app uses the Hexagon NPU delegate for fastest inference. On Google Pixel 10 (Tensor G5) the app uses `Backend.GPU` (ARM Immortalis-G925 via LiteRT). Performance and output are equivalent — the hardware tier detection automatically selects the right delegate. See [`models/README.md`](models/README.md) for per-device model setup.
+> **Backend note:** On Qualcomm devices the app uses the GPU (OpenCL) delegate via LiteRT. Hardware tier detection automatically selects the right delegate. See [`models/README.md`](models/README.md) for per-device model setup.
+
+### Performance — Samsung Galaxy S23 Ultra (Gemma-4 E-4B, GPU)
+
+Measured on-device from production logs. "tok/s" counts output tokens delivered to the UI.
+
+| Metric | Value |
+|--------|-------|
+| Time to First Token (TTFT) | 2–5s (cold), ~2s (warm KV cache) |
+| Generation speed | ~9–10 tok/s |
+| Engine init | ~2s (warm, kernel cache hit) |
+
+**Multi-Token Prediction (MTP / speculative decoding)**
+
+MTP uses a bundled drafter model to speculatively generate multiple tokens per GPU step. The benefit scales with response length — on long responses (~400+ tokens) we observed a **~2× wall-time speedup**; on short responses the overhead is negligible. Toggle in Settings → Model → E-4B panel.
+
+| Response length | MTP off | MTP on | Speedup |
+|----------------|---------|--------|---------|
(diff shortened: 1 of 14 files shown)


lokhor:
## Code Review — `kai-772-mtp`

Four issues found, two critical.

---

### 🔴 Critical — Missing Room migration for `speculativeDecodingEnabled`

`speculativeDecodingEnabled` is added to the `model_settings` entity but the database version is not incremented. Room validates the identity hash against the live schema on startup — every existing install will crash on upgrade with `IllegalStateException: Room cannot verify the data integrity` because the `model_settings` table on disk is missing the new column. There is no `fallbackToDestructiveMigration` in `MemoryModule`, so this is a hard crash with no recovery path.

**Fix:** Bump `@Database(version = 31)`, add schema file `31.json`, and register a new migration:

```kotlin
val MIGRATION_30_31 = object : Migration(30, 31) {
    override fun migrate(db: SupportSQLiteDatabase) {
        db.execSQL(
            "ALTER TABLE model_settings ADD COLUMN speculativeDecodingEnabled INTEGER NOT NULL DEFAULT 0"
        )
    }
}
```

Register it in `MemoryModule.addMigrations(...)`.

---

### 🔴 High — Delete button exposed on required models

The old code gated the Delete button behind `if (!model.isRequired)`. This PR removes that guard entirely when adding the Update button, so users can now tap Delete on E2B and EmbeddingGemma (both `isRequired = true`). The active inference session immediately loses its model file — any in-progress or subsequent generation attempt will crash the engine.

**Note on the intended behaviour:** E2B and EmbeddingGemma are required and must support Update (so users can get a newer version). They must NOT support Delete (removing them leaves the app broken until a full re-download completes). The Update button should be independent of `isRequired`.

**Fix:** Restore the guard for Delete only, leaving Update ungated:

```kotlin
// Update — available for all models including required ones
TextButton(onClick = onUpdate) {
    Text("Update")
}
// Delete — only for optional models
if (!model.isRequired) {
    Spacer(modifier = Modifier.width(4.dp))
    TextButton(onClick = onDelete) {
        Text("Delete", color = MaterialTheme.colorScheme.error)
    }
}
```

---

### 🟠 High — FLAGSHIP fresh installs now auto-download 5.8 GB instead of 3.4 GB

The removed filter condition was the only thing preventing the required-model auto-queue loop from queuing E2B (2.4 GB) on FLAGSHIP devices. With it gone, both E2B and E4B are queued simultaneously on first launch — E2B is never actually used on FLAGSHIP (E4B is the active model). This wastes 2.4 GB of bandwidth and storage on every new FLAGSHIP install.

The auto-queue exclusion and the Update button in `ModelManagementScreen` are independent mechanisms — removing the exclusion is not needed to enable user-triggered updates.

**Fix:** Restore the FLAGSHIP exclusion in the required-model auto-queue loop:

```kotlin
it.isRequired && !it.isDownloaded(context) &&
(!it.isGated || authRepository.getAccessToken() != null) &&
!(it == KernelModel.GEMMA_4_E2B && tier == HardwareTier.FLAGSHIP)
```

---

### 🟡 Medium — `_speculativeDecodingEnabled` StateFlow is dead code

`private val _speculativeDecodingEnabled = MutableStateFlow(false)` in `ChatViewModel` is declared and written to but never read or exposed. The flag reaches the engine correctly via `settings.speculativeDecodingEnabled` passed directly into `ModelConfig` — this field is redundant.

**Fix:** Remove the declaration and the two assignment lines.

---

### ✅ No issues with

- Speculative decoding flag lifecycle (`buildConversationConfig` → `resetExperimentalFlags` pattern is correct)
- `startDownload(force = true)` path in `ModelManagementViewModel` (correctly uses `REPLACE` policy)
- `ExperimentalFlags` thread safety (all calls inside `withContext(LlmDispatcher)`)

PR #816, opened 2026-05-09
feat: add deterministic unit conversion intent

### LINKED ISSUE #676: unit conversion tool
Convert between units tool to use in quick actions

Grams to kilograms
Grams to ounces 
Kilograms to pounds

Miles to kilometers 
Mph to kmh

And so on

lokhor:
## Summary
- add a native `convert_units` intent backed by a deterministic `UnitConversionEvaluator` for common unit families: length, mass, volume, temperature, and speed
- route common conversion phrasing through QuickIntentRouter and teach `run_intent` to call the conversion tool instead of leaving those requests to model math
- keep ingredient-aware cooking density conversions out of this PR; those stay in follow-up issue #827
- mark merged calculator work as done in `docs/ROADMAP.md`

## Verification
- `./gradlew :core:skills:testDebugUnitTest --tests "*UnitConversionEvaluatorTest" --tests "*QuickIntentRouterTest" --tests "*NativeIntentHandlerTest" :core:skills:compileDebugKotlin`
- `./gradlew :app:compileDebugKotlin`

### Routing and supported families
- [ ] Ask `100m to yards` and confirm it routes through the native conversion intent and returns an approximate yard value.
- [ ] Ask `convert 1.5 m to cm` and confirm exact metric scaling returns `150 cm` without an approximate qualifier.
- [ ] Ask `convert 500 grams to kg` and confirm exact metric mass scaling returns `0.5 kg`.
- [ ] Ask `convert 1 gallon to liters` and confirm the deterministic result matches the native tool rather than an LLM-estimated answer.
- [ ] Ask `convert 60 mph to km/h` and confirm the reply is deterministic.
- [ ] Ask `60 mph in m/s` and confirm the speed conversion is supported end-to-end.
- [ ] Ask `how many cups are in 2 liters` and confirm reversed phrasing still routes through the native conversion intent.
- [ ] Ask `convert 100m to yards` from Quick Actions voice and confirm the same native conversion path is used there too.

### Temperature behavior
- [ ] Ask `32 fahrenheit to celsius` and confirm it returns `0 degrees Celsius` without failing on a negative intermediate.
- [ ] Ask `convert -40 celsius to fahrenheit` and confirm valid negative temperatures are supported.
- [ ] Ask `convert 0 celsius to kelvin` and confirm it returns `273.15 kelvin`.
- [ ] Ask `convert 1 fahrenheit to celsius` and confirm the spoken/displayed answer is treated as approximate rather than falsely exact.

### Error handling and boundaries
- [ ] Ask `convert -5 miles to km` and confirm the failure is clear and non-crashing.
- [ ] Ask `convert -500 fahrenheit to celsius` and confirm values below absolute zero fail clearly.
- [ ] Ask `convert 5 miles to kilograms` and confirm cross-category conversions fail clearly instead of fabricating an answer.
- [ ] Ask `convert 5 miles to parsecs` and confirm unsupported target units fail clearly.
- [ ] Ask an ingredient-aware cooking query like `how much does 3 tbsp of butter weigh` and confirm it is still left to the separate cooking-conversion follow-up rather than pretending generic units solve it.

Closes #676


core/skills/src/main/java/com/kernel/ai/core/skills/natives/UnitConversionEvaluator.kt
```diff
@@ -0,0 +1,164 @@
...
+            fromUnit = fromUnit,
+            toUnit = toUnit,
+            isApproximate = converted.second,
+        )
+    }
+
+    private fun parseValue(rawValue: String): BigDecimal {
+        val trimmed = rawValue.trim()
+        require(trimmed.isNotBlank()) { "No conversion value provided" }
+        require(trimmed.length <= MAX_INPUT_LENGTH) { "Conversion value is too long" }
```
lokhor, line 358: **Bug — negative values produce nonsensical output**

`parseValue` accepts negative inputs because the regex uses `-?\\d+` and `BigDecimal` parses negatives silently. "convert -5 miles to km" routes through QIR, reaches `convertUnits`, and returns "-5 miles is -8.04672 kilometers." which gets shown/spoken to the user.

Add a guard immediately after parsing:
```kotlin
return try {
    BigDecimal(trimmed).also { v ->
        require(v >= BigDecimal.ZERO) { "Conversion value must be non-negative" }
    }
} catch (_: NumberFormatException) {
    throw IllegalArgumentException("Invalid conversion value '$rawValue'")
}
```
Also add a corresponding test case in `UnitConversionEvaluatorTest`.

core/skills/src/main/java/com/kernel/ai/core/skills/natives/UnitConversionEvaluator.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private fun divideWithApproximation(left: BigDecimal, right: BigDecimal): Pair<BigDecimal, Boolean> =
+        try {
+            left.divide(right) to false
+        } catch (_: ArithmeticException) {
+            left.divide(right, APPROXIMATE_SCALE, RoundingMode.HALF_UP) to true
+        }
+
+    private fun normalizeUnit(raw: String): String =
```
lokhor, line 442: **Correctness — `km / h` (spaced slash) silently fails alias lookup**

`normalizeUnit` collapses dots to spaces but leaves `/` untouched, so if the LLM sends `from_unit="km / h"` (with surrounding spaces), the normalised form is `"km / h"` which has no entry in the alias map → `SupportedUnit.from` returns `null` → `IllegalArgumentException("Unsupported unit 'km / h'")`. The QIR always emits `"km/h"` (no spaces), but the LLM-path through `RunIntentSkill` is free to produce any casing/spacing.

Fix: strip spaces around `/` in normalisation:
```kotlin
private fun normalizeUnit(raw: String): String =
    raw.lowercase(Locale.ENGLISH)
        .replace('.', ' ')
        .replace(Regex("""\\s*/\\s*"""), "/")   // "km / h" → "km/h"
        .replace(Regex("""\\\s+"""), " ")
        .trim()
```

core/skills/src/main/java/com/kernel/ai/core/skills/natives/NativeIntentHandler.kt
```diff
@@ -1980,6 +1982,28 @@ class NativeIntentHandler @Inject constructor(
...
+        return try {
+            val result = UnitConversionEvaluator.convert(value, fromUnit, toUnit)
+            val sourceDisplay = result.fromUnit.displayName(result.inputValue)
+            val targetDisplay = result.toUnit.displayName(result.outputValue)
+            val content = if (result.isApproximate) {
+                "${result.inputValue.toPlainString()} $sourceDisplay is approximately ${result.outputValue.toPlainString()} $targetDisplay."
+            } else {
+                "${result.inputValue.toPlainString()} $sourceDisplay is ${result.outputValue.toPlainString()} $targetDisplay."
+            }
+            SkillResult.DirectReply(content)
```
lokhor: **Voice UX — approximate results with 8 decimal places are TTS-hostile**

For an approximate conversion like 5 kg → pounds, `content` is `"5 kilograms is approximately 11.02311311 pounds."`. TTS will read every digit: *"eleven point zero two three one one three one one"* — jarring in voice mode.

`SkillResult.DirectReply` accepts an optional `spokenSummary`. Set it to a version rounded to 2 dp for voice surfaces:
```kotlin
val spoken = if (result.isApproximate) {
    val rounded = result.outputValue.setScale(2, RoundingMode.HALF_UP).stripTrailingZeros()
    "${result.inputValue.toPlainString()} $sourceDisplay is approximately ${rounded.toPlainString()} $targetDisplay."
} else {
    content
}
SkillResult.DirectReply(content, spokenSummary = spoken)
```
Exact conversions (e.g. 8.04672 km) are short enough that TTS handles them fine; only approximate results need the rounded form.

feature/chat/src/test/java/com/kernel/ai/feature/chat/ChatViewModelVoiceTest.kt
```diff
@@ -204,7 +204,7 @@ class ChatViewModelVoiceTest {
         coVerify {
             voiceOutputController.openStreamingSession(any())
         }
-        assertEquals(listOf("Hi back" to true), streamedChunks)
+        assertEquals(listOf("Hi back." to true), streamedChunks)
```
lokhor, line 207: **Test assertion changed from PR #818 — needs explanation**

PR #818 introduced this test with `"Hi back"` (no trailing period). This commit changes it to `"Hi back."`. The generated token at line 118 is `GenerationResult.Token("Hi back")` (no period), so the period is being added somewhere in the streaming/finalisation path.

If `finalizeChatTextForSpeech` was changed to append a sentence-ending period when none is present, that's a behaviour change that belongs in the commit message and should have a dedicated test. If the test expectation in #818 was simply wrong all along, please call that out explicitly.

Same question applies to line 339.

lokhor (commented):
**No conflict with PR #818.** The branch was cut from `main` *after* #818 merged (both commits sit on top of `ba701a6c`). The `ChatViewModel.speakMessage()` changes in commit `c53cfe7c` build cleanly on top of the `openStreamingSession()` path — they add error handling and fix speaker-button state tracking, not replace or regress the streaming path.

Four issues worth addressing before merge:
1. **Bug** — negative values reach the conversion and produce nonsensical output ("-5 miles is -8.04672 kilometers.")
2. **Voice UX** — approximate results with 8 decimal places are TTS-hostile; `spokenSummary` should be set to a rounded form
3. **Correctness** — `normalizeUnit` normalises `.→space` but not `/`; "km / h" (space-padded slash) silently fails alias lookup
4. **Test explanation needed** — two assertions inherited from PR #818 are changed from `"Hi back"` to `"Hi back."` without explanation in the commit message

lokhor:
Targeted manual tests to run on this build:

- Voice/chat routing
  - `How many cups in 2 L` → should route natively to `convert_units`, not `llm_fallthrough`.
  - `How many yards in 350 m` → should route natively to `convert_units`, not `llm_fallthrough`.
  - `Convert 100 km an hour to metres a second` → should route natively to `convert_units`, not `llm_fallthrough`.

- Mixed-unit conversions
  - `Convert 189 cm to feet and inches` → should return mixed output in feet + inches.
  - `Convert 6 feet 2 inches to cm` → should route natively and return `187.96 centimeters`.

- Regression checks
  - `Convert 1 gallon to liters` → should still return the deterministic exact reply, not fall through.
  - Quick Actions voice: `100m to yards` → should use the same native conversion path as chat.

Please grab logs for any miss that still lands in `llm_fallthrough` so we can add the exact spoken variant.

lokhor:
Additional targeted manual check after the latest push:

- `Convert 1 gallon to litres`
  - should still route to `convert_units`
  - display text should remain full precision: `1 gallon is 3.785411784 liters.`
  - spoken/TTS reply should now be rounded: `1 gallon is 3.79 liters.`

Related fix in this push: exact conversions with long decimal outputs now emit a rounded `spokenSummary` instead of making TTS read the full display precision.

PR #825, opened 2026-05-09
feat: start-listening audio cue + slot-fill retry (#791, #790)

### LINKED ISSUE #790: Voice slot-fill should retry on no-speech and support cancel phrases
## Summary
During Quick Actions multi-slot voice flows, a slot reply can fail with a raw error when no speech is heard or understood. The flow should handle this as a recoverable interaction state instead of surfacing an opaque failure.

## Problem
- When a user is in an active multi-slot Quick Actions flow and their spoken reply is not captured, the app can show an error instead of reprompting.
- There is no explicit voice escape path for phrases like `stop`, `cancel`, or `stop that` while slot filling is active.
- We should distinguish active slot-fill reply capture from the initial idle mic tap, because retry behavior should likely be more conservative when the user has not yet committed to a specific slot prompt.

### Active slot-fill reply sessions
- On `ERROR_NO_MATCH` / no understood speech / timeout, automatically reprompt up to **2 times**.
- The reprompt should be explicit and brief, e.g. "Sorry, I didn’t catch that. Please repeat that."
- After the retry budget is exhausted, keep the slot prompt visible and wait for manual input instead of looping indefinitely.
- Recognize escape phrases such as:
  - `stop`
  - `cancel`
  - `stop that`
  - any existing app-standard cancel grammar if broader voice cancellation already exists elsewhere
- On cancel, abort the slot-fill flow cleanly and reset transient voice state.

### Initial Quick Actions mic press from idle
- If we add retry here at all, keep it to **at most 1 automatic retry**.
- Rationale: repeated auto-retries from idle are more likely to feel noisy or confusing than helpful.

## Acceptance criteria
- Multi-slot Quick Actions voice replies retry up to 2 times on no-speech / no-match.
- Cancel/stop phrases can abort slot filling at any step.
- Idle mic-start behavior is handled separately from active slot-reply behavior.
- No infinite retry loops.
- Add regression coverage for retry-budget exhaustion and cancel handling.

## Notes
This came from device testing while validating PR #787 follow-ups.

### LINKED ISSUE #791: Add a start-listening cue for Quick Actions voice capture
## Summary
Quick Actions voice capture currently has no strong start-listening cue after the mic button is pressed. In device use, it is easy to start speaking before recognition is actually ready.

## Problem
- Users can begin speaking before the device is listening.
- This leads to clipped utterances and avoidable no-match errors.
- The Quick Actions voice flow needs a clearer transition into active listening.

## Proposed behavior
- Add a short, unobtrusive start-listening cue when Quick Actions voice capture is actually ready.
- A brief earcon / "bloop" is the primary candidate.
- The cue should align with the real listening state, not just button press, so users are trained on a truthful signal.

## Acceptance criteria
- Quick Actions voice capture emits a clear start-listening cue when the recognizer is ready.
- The cue is not played early on button press if listening has not started yet.
- The cue does not regress existing voice flows or create duplicate audio prompts.
- Add targeted verification for the listening-state transition that triggers the cue.

## Notes
This came from device testing during PR #787 validation.

lokhor:
## Summary

Implements two Quick Actions voice UX improvements:

- **#791** — Start-listening audio cue: a short beep plays the instant the mic opens (aligned to `ListeningStarted` event, not button press)
- **#790** — Slot-fill retry on no-speech: up to 2 reprompts on voice error during slot-fill, cancel phrase escape, 1 retry for idle command mode

### #791 — Start-listening cue
- `StartListeningCuePlayer` interface + `ToneStartListeningCuePlayer` impl (`ToneGenerator.TONE_PROP_BEEP`, 100ms, `STREAM_MUSIC` at 60% volume)
- Injected into `ActionsViewModel`; `playCue()` called in `ListeningStarted` handler, guarded by `ownsVoiceCapture()`
- Hilt binding in `VoiceModule`

### #790 — Slot-fill retry + cancel escape
- SlotReply errors: up to 2 retries with reprompt `"Sorry, I didn't catch that. <prompt>"`; on exhaustion, slot card stays for manual input
- Command mode errors: 1 retry, then surface error
- Retry budgets reset per slot (`primePendingSlot`) / per press (`startVoiceCommand`)
- Escape phrases `"stop"` / `"cancel"` / `"stop that"` in SlotReply → `cancelSlotFill()` immediately

## Tests

9 new unit tests in `ActionsViewModelVoiceTest`:
- Cue fires on `ListeningStarted` (owned mode only)
- Slot-fill retries ×2 then waits for manual input
- Cancel/stop/stop-that phrases abort slot-fill
- Idle command retries once, resets on fresh press

## Pre-existing failures (not introduced here)
- `ChatViewModelVoiceTest` ×2 and `LatexConversionTest` — confirmed identical on `main` before any changes

## Testing on device
- [ ] Press mic in Quick Actions → short beep plays when recogniser is ready (not on press)
- [ ] Speak nothing during a slot prompt → reprompt plays, up to 2 times, then falls back to text input
- [ ] Say "cancel" or "stop" during slot prompt → flow aborts cleanly

Closes #791
Closes #790

### DIFF
--- a/core/voice/src/main/java/com/kernel/ai/core/voice/ToneStartListeningCuePlayer.kt
+++ b/core/voice/src/main/java/com/kernel/ai/core/voice/ToneStartListeningCuePlayer.kt
@@ -1,0 +1,40 @@
+package com.kernel.ai.core.voice
+
+import android.media.AudioManager
+import android.media.ToneGenerator
+import android.util.Log
+import javax.inject.Inject
+import javax.inject.Singleton
+
+private const val TAG = "KernelAI"
+private const val CUE_VOLUME_PERCENT = 60
+private const val CUE_DURATION_MS = 100
+
+/**
+ * Plays a brief, non-jarring beep via [ToneGenerator] when the microphone
+ * becomes ready for speech.
+ *
+ * Uses [ToneGenerator.TONE_PROP_BEEP] on [AudioManager.STREAM_MUSIC] at 60%
+ * volume for a 100 ms duration — a short, clean UI earcon that mirrors the
+ * convention used by [LoopListeningCue].
+ */
+@Singleton
+class ToneStartListeningCuePlayer @Inject constructor() : StartListeningCuePlayer {
+
+    private val toneGenerator by lazy {
+        try {
+            ToneGenerator(AudioManager.STREAM_MUSIC, CUE_VOLUME_PERCENT)
+        } catch (e: RuntimeException) {
+            Log.w(TAG, "ToneStartListeningCuePlayer: failed to create ToneGenerator", e)
+            null
+        }
+    }
+
+    override fun playCue() {
+        try {
+            toneGenerator?.startTone(ToneGenerator.TONE_PROP_BEEP, CUE_DURATION_MS)
+        } catch (e: RuntimeException) {
+            Log.w(TAG, "ToneStartListeningCuePlayer: failed to play cue", e)
+        }
+    }
+}
--- a/core/voice/src/main/java/com/kernel/ai/core/voice/ToneStartListeningCuePlayer.kt
+++ b/core/voice/src/main/java/com/kernel/ai/core/voice/ToneStartListeningCuePlayer.kt
@@ -21,20 +21,46 @@ private const val CUE_DURATION_MS = 100
 @Singleton
 class ToneStartListeningCuePlayer @Inject constructor() : StartListeningCuePlayer {
 
-    private val toneGenerator by lazy {
-        try {
-            ToneGenerator(AudioManager.STREAM_MUSIC, CUE_VOLUME_PERCENT)
-        } catch (e: RuntimeException) {
-            Log.w(TAG, "ToneStartListeningCuePlayer: failed to create ToneGenerator", e)
-            null
+    /**
+     * Held for process lifetime as a @Singleton. [release] can be called to free the native
+     * AudioTrack if the audio system enters a bad state — it is not called automatically on
+     * ViewModel teardown because this singleton outlives any single ViewModel.
+     */
+    private var toneGeneratorInstance: ToneGenerator? = null
+    private val toneGenerator: ToneGenerator?
+        get() {
+            if (toneGeneratorInstance == null) {
+                toneGeneratorInstance = try {
+                    ToneGenerator(AudioManager.STREAM_MUSIC, CUE_VOLUME_PERCENT)
+                } catch (e: RuntimeException) {
+                    Log.w(TAG, "ToneStartListeningCuePlayer: failed to create ToneGenerator", e)
+                    null
+                }
+            }
+            return toneGeneratorInstance
         }
-    }
 
     override fun playCue() {
         try {
             toneGenerator?.startTone(ToneGenerator.TONE_PROP_BEEP, CUE_DURATION_MS)
         } catch (e: RuntimeException) {
             Log.w(TAG, "ToneStartListeningCuePlayer: failed to play cue", e)
+        }
+    }
+
+    /**
+     * Releases the native [ToneGenerator] AudioTrack resource and resets the internal reference
+     * so that the next [playCue] call will recreate it.
+     *
+     * Call this when the audio system enters a bad state and needs to be reset.
+     */
+    override fun release() {
+        try {
+            toneGeneratorInstance?.release()
+        } catch (e: RuntimeException) {
+            Log.w(TAG, "ToneStartListeningCuePlayer: failed to release ToneGenerator", e)
+        } finally {
+            toneGeneratorInstance = null
         }
     }
 }
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsViewModel.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsViewModel.kt
@@ -262,7 +262,16 @@ class ActionsViewModel @Inject constructor(
                                             "$commandVoiceRetryCount/$COMMAND_MAX_VOICE_RETRIES",
                                     )
                                     _voiceCaptureState.value = VoiceCaptureState.Idle
-                                    startVoiceCapture(VoiceCaptureMode.Command)
+                                    // Speak reprompt then re-open mic after TTS completes.
+                                    // If spoken responses are off, re-open immediately.
+                                    pendingVoiceSpeechJob = viewModelScope.launch {
+                                        if (spokenResponsesEnabled) {
+                                            voiceOutputController.speak(
+                                                VoiceSpeakRequest("Sorry, I didn't catch that. Please try again.")
+                                            )
+                                        }
+                                        startVoiceCapture(VoiceCaptureMode.Command)
+                                    }
                                 } else {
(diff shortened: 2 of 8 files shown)


core/voice/src/main/java/com/kernel/ai/core/voice/ToneStartListeningCuePlayer.kt
```diff
@@ -0,0 +1,40 @@
...
+            ToneGenerator(AudioManager.STREAM_MUSIC, CUE_VOLUME_PERCENT)
+        } catch (e: RuntimeException) {
+            Log.w(TAG, "ToneStartListeningCuePlayer: failed to create ToneGenerator", e)
+            null
+        }
+    }
+
+    override fun playCue() {
+        try {
+            toneGenerator?.startTone(ToneGenerator.TONE_PROP_BEEP, CUE_DURATION_MS)
```
lokhor, line 45: 🟡 **Medium — `ToneGenerator` is never released**

`ToneGenerator` wraps a native `AudioTrack` and its docs require `release()` to be called when it's no longer needed. The `lazy` block creates it once and holds it forever with no cleanup path.

Since this is `@Singleton`, the instance lives for the entire process — not a growing leak, but the native resource is silently held across the whole app lifetime. If the audio system ever puts `ToneGenerator` into a bad state and needs a reset, the `lazy` guard means you can't recover without a process restart.

**Fix:** Implement `Closeable` (or add a `release()` method) and call `toneGenerator?.release()` from it. Wire it to an Application-scoped lifecycle callback or expose it on the `StartListeningCuePlayer` interface if teardown is ever needed:
```kotlin
fun release() {
    toneGenerator?.release()
}
```
At minimum, document that the resource is intentionally held for app lifetime so the next reader doesn't have to rediscover this.

feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsViewModel.kt
```diff
@@ -267,6 +345,8 @@ class ActionsViewModel @Inject constructor(
     // ── Public API ──────────────────────────────────────────────────────────
 
     fun startVoiceCommand() {
+        // #790: Reset command retry budget on fresh user-initiated press.
+        commandVoiceRetryCount = 0
```
lokhor, line 359: `commandVoiceRetryCount = 0` is correctly reset here on every user-initiated press. `startVoiceSlotReply()` (line 353) is missing the equivalent `slotReplyVoiceRetryCount = 0` reset — see the overall review body for the full failure scenario.

lokhor (commented):
## Review — feat: start-listening audio cue + slot-fill retry (#791, #790)

One real bug and one resource-management issue. See inline comments for specifics.

---

### 🔴 High — `startVoiceSlotReply()` doesn't reset `slotReplyVoiceRetryCount`

**File:** `feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsViewModel.kt` — `startVoiceSlotReply()` (line 353)

`startVoiceCommand()` correctly resets `commandVoiceRetryCount = 0` on every user-initiated press. `startVoiceSlotReply()` has no equivalent reset — `slotReplyVoiceRetryCount` is only zeroed in `primePendingSlot()`, which is not called when the user manually presses the mic button on a slot card that is *already* showing.

**Failure scenario:**
1. Slot card opens (voice mode). `slotReplyVoiceRetryCount = 0` ✅
2. Mic picks up nothing → 2 auto-retries exhaust the budget (`count = 2`).
3. Slot card stays visible for manual input.
4. User taps mic again — `startVoiceSlotReply()` is called. `slotReplyVoiceRetryCount` is still 2.
5. Voice error → `2 >= SLOT_REPLY_MAX_VOICE_RETRIES (2)` → falls straight to the exhausted branch. Zero retries for this session. The mic button silently does nothing useful.

**Fix:**
```kotlin
fun startVoiceSlotReply() {
    if (_pendingSlot.value == null) return
    Log.d(TAG, "ActionsViewModel: startVoiceSlotReply pendingSlot=true")
    slotReplyVoiceRetryCount = 0   // ← add this, mirrors startVoiceCommand()
    setSlotReplyAutoRearmArmed(false, "startVoiceSlotReply")
    clearExpectedSlotPromptSpeech()
    startVoiceCapture(VoiceCaptureMode.SlotReply)
}
```
Also add a test: exhaust the budget, tap mic again, emit one error, assert `pendingSlot` is still non-null and a retry TTS was spoken.

lokhor:
## Scoped re-review — fix commits 50a245c & d33c146

Reviewed both fix commits only. Findings below.

---

### Fix 1 — `slotReplyVoiceRetryCount` reset (`50a245c`)

**Correct.** The reset is placed after the `if (_pendingSlot.value == null) return` guard, so it only fires when a real slot session is being started — identical structure to the `commandVoiceRetryCount = 0` reset in `startVoiceCommand()`. Resetting to 0 on a first tap (where the count is already 0) is harmless.

No regressions: `primePendingSlot()` still resets on new intent routing; `startVoiceSlotReply()` now also resets on manual re-tap. The two reset sites cover all entry paths.

**Test — substantive, not trivially passing.**

The test correctly:
1. Exhausts the full 2-retry budget event-by-event (including TTS reprompt + auto-rearm cycles), then asserts `exactly = 2` speaks and `VoiceCaptureState.Idle` — catching any over-triggering during exhaustion.
2. Calls `startVoiceSlotReply()` again (the manual re-tap under test), emits one more error, and asserts `atLeast = 3` total speaks.

The `atLeast = 3` assertion is the right choice: `exactly = 3` would be brittle if the auto-rearm cycle in the new session triggers another speak before the test ends. The assertion confirms the retry path fired (not the exhaustion path) without over-constraining the count.

---

### Fix 2 — `ToneGenerator` nullable lifecycle (`d33c146`)

**No NPE paths.** The `toneGenerator` computed getter returns `ToneGenerator?`; `playCue()` accesses it with `?.startTone(...)` — same nullability as the original `by lazy` path.

**No double-release risk.** `release()` uses `finally { toneGeneratorInstance = null }` so the reference is cleared even if `toneGeneratorInstance.release()` throws. A second call finds `null`, skips the try body, and sets null again — clean.

**Recreation after release works correctly.** The getter's `if (toneGeneratorInstance == null)` check causes a fresh `ToneGenerator` to be constructed on the next `playCue()` call after a reset.

**One observation — Medium, low actual risk:**

`by lazy` defaults to `LazyThreadSafetyMode.SYNCHRONIZED`, so the original code had an implicit lock on first access. The new `private var` + computed getter has no synchronization. If `release()` were ever called from a background thread concurrently with `playCue()`, the double-checked pattern in the getter could read a stale `null` or race on assignment.

In practice this is unlikely — `playCue()` is triggered by `ListeningStarted` events processed on the main thread via `viewModelScope`, and `release()` is documented as a manual recovery hook (also main-thread). The KDoc makes the intent clear. No action required unless a future caller invokes `release()` off-thread, at which point adding `@Volatile` to `toneGeneratorInstance` would close it.

---

### Summary

Both fixes are correct. The test exercises the real failure case end-to-end and is not trivially passing. The only observation is the implicit loss of `by lazy`'s synchronized semantics in Fix 2, which carries negligible risk given the documented main-thread usage pattern. No blockers.

PR #946, opened 2026-05-20
fix(#941): pass enable_thinking extraContext to activate Gemma thinking mode

### LINKED ISSUE #694: Thinking Mode
I think we should review how edge gallery handles thinking mode as the current implementation seemingly does nothing:

https://github.com/google-ai-edge/gallery/blob/3b368f4aefec2358bdd915e3577e86dfcc5ba9d8/Android/src/app/src/main/proto/settings.proto#L61

https://github.com/google-ai-edge/gallery/blob/3b368f4aefec2358bdd915e3577e86dfcc5ba9d8/Android/src/app/src/main/java/com/google/ai/edge/gallery/ui/common/chat/MessageBodyThinking.kt#L4

### LINKED ISSUE #941: research: investigate thinking mode on public Gemma-4 E-4B — are thinking tokens actually emitted?
## Summary

There is no visible evidence that Gemma-4 E-4B is emitting chain-of-thought (thinking) tokens in the app, despite our infrastructure being fully wired up. This issue tracks the investigation and any resulting fix.

## What we have

Our `LiteRtInferenceEngine` correctly registers the thought channel via:

```kotlin
listOf(Channel("thought", "<|think|>", "<|/think|>"))
```

And routes thinking tokens via:

```kotlin
val thinkingText = message.channels["thought"]
if (!thinkingText.isNullOrEmpty()) {
    trySend(GenerationResult.Thinking(thinkingText))
}
```

The UI layer handles `GenerationResult.Thinking` — collapsible thinking bubbles render in the chat, and `showThinkingProcess` is a toggleable setting in Model Settings. The code path is complete.

## The suspected problem

The **public Gemma-4 E-4B `.litertlm` file** (downloaded from HuggingFace) may not be the thinking-enabled model variant. We know from prior research (issue #690) that Google AI Edge Gallery uses an internal build (`gemma4_4b_v09_obfus_fix_all_modalities_thinking.litertlm`) that is **not publicly available**. If the public model binary does not emit `<|think|>` … `<|/think|>` tokens, our channel registration is a no-op — it registers a channel that is never written to.

Edge Gallery marks models with `ModelCapability.LLM_THINKING` to guard thinking-mode UI on models that actually support it. We don't have this guard.

## Investigation tasks

1. **Verify with logcat** — Enable thinking and send a reasoning-heavy prompt (e.g. "What is 37 × 48? Work through it step by step."). Check `KernelAI` logcat for "Thinking tokens: N chars". If `N == 0` consistently, the model is not emitting thinking tokens.

2. **Check the model file** — Compare the SHA256 of our downloaded E4B `.litertlm` against any published checksum. Identify which HuggingFace commit/variant we are using.

3. **Search for a publicly available thinking variant** — Check:
   - `google/gemma-4-4b-it-litert-preview` on HuggingFace for any `thinking` variant
   - The LiteRT GitHub and community channels for a public thinking-enabled E4B build
   - Whether Gemma-4 E-2B has a public thinking variant that could be used as a test

4. **Test with an explicit thinking prompt prefix** — Some thinking-enabled models require the thinking section to be explicitly initiated in the prompt template. Test if prepending `<|think|>` to the user prompt has any effect.

5. **Review ConversationConfig channels documentation** — Confirm that `Channel("thought", "<|think|>", "<|/think|>")` is the correct token pair for the public E4B model (not an internal-only variant of the delimiters).

## Expected outcome

- Confirm whether the public E4B emits thinking tokens.
- If not: document the limitation; gate the thinking UI toggle behind a model capability check so users aren't misled.
- If yes: identify what prompt or config change is needed to trigger it.
- If a public thinking variant exists: update the model download URL and SHA256 in our model config.

## Acceptance criteria

- [ ] Logcat test confirms presence or absence of thinking tokens for public E4B
- [ ] If thinking does not work: thinking toggle in Model Settings is hidden/disabled for models that don't support it, with a short explanation
- [ ] If thinking works: provide a test prompt that reliably triggers visible chain-of-thought output
- [ ] If a public thinking-enabled model is found: update model config and SHA256; confirm thinking bubbles render on device

## Related

- #690 — Edge Gallery vs our E4B model package differences
- #704 — Phase 3A: Model, runtime and inference experiments

### LINKED ISSUE #953: Refactor thinking-mode streaming to a buffered state machine
Parent epic: #963 — Chat UX overhaul and multimodal composer foundation

## Summary

Thinking-mode streaming on LiteRT-LM is still too brittle because the SDK does not emit one stable callback shape across turns. We fixed the immediate truncation bug in PR #946, but the investigation showed the parser should be simplified into a buffered state machine instead of continuing to grow callback-shape heuristics.

## Key findings from the investigation

1. LiteRT can emit the thinking/response boundary inconsistently:
   - sometimes it appears inside `message.channels["thought"]`
   - sometimes it appears only in `message.toString()`
   - sometimes the marker appears visually split across lines
   - sometimes post-close visible response text continues arriving in `channels["thought"]` while `message.toString()` is empty
2. The same model can behave differently on later turns in the same chat.
3. A cleaner UI model like Edge Gallery's separate `ChatMessageThinking` is useful, but it does not solve the root problem by itself because the bug happens in the streaming/parser layer before UI rendering.

## Proposed follow-up

Replace the current branch-heavy per-callback routing with a buffered state machine that only runs when thinking mode is enabled for the current generation.

### Suggested behavior

1. **Pre-close state**
   - Buffer incoming content from both `message.channels["thought"]` and `message.toString()`.
   - Emit only thinking content before the close marker is observed.
2. **Close detection**
   - Detect it even when it is split across callback boundaries or includes line breaks/whitespace.
   - On first detection, split buffered content exactly once into:
     - thinking prefix
     - visible response suffix
3. **Post-close state**
   - After the split, route all subsequent content to the visible response stream regardless of whether it arrives via `channels["thought"]` or `message.toString()`.
   - Keep duplicate suppression, but make it conservative and state-aware.
4. **Non-thinking generations**
   - Preserve the current non-thinking streaming path as much as possible.

## Why this is worth doing

The current implementation now covers several traced SDK quirks, but each new callback shape has required another special-case branch. A state-machine parser should be easier to reason about, more robust across turns, and less likely to regress when LiteRT changes callback behavior.

## Nice-to-have follow-up after parser stabilization

Consider a separate chat-level refactor toward an Edge Gallery-style thinking message model:

- dedicated thinking message type
- explicit in-progress/done thinking state
- cleaner UI and persistence boundaries between hidden reasoning and visible assistant text

That should be treated as a follow-up to parser stabilization, not a substitute for it.

## Related

- PR #946
- Issue #941
- Edge Gallery reference commit: `9a5e1930264e19a125bb4af875d1105e14bcbe54`

### LINKED ISSUE #955: docs: Edge Gallery thinking mode implementation — full analysis
## Summary

This documents **how Google AI Edge Gallery implements thinking mode end-to-end** — from model capability declaration through inference, response parsing, and UI rendering. The analysis was done by reading the Edge Gallery source code (`google-ai-edge/gallery` repo).

The goal is to give us a concrete reference for our own implementation, especially for the UI rendering layer (collapsible thinking bubbles) and the model capability gating pattern.

---

### 1. Model Capability Gating

Edge Gallery uses a `ModelCapability.LLM_THINKING` enum value that must be declared in the model's allowlist JSON entry. The thinking toggle only appears in settings if the model declares this capability.

```json
{
  "capabilities": ["llm_thinking", "speculative_decoding"],
  "capabilityToTaskTypes": {
    "llm_thinking": ["llm_chat", "llm_ask_image", "llm_ask_audio"]
  }
}
```

**Relevant for us:** We should gate the `showThinkingProcess` toggle behind a similar capability check so users aren't misled when using a model that doesn't emit thinking tokens.

### 2. Response Channel Extraction

The thinking content comes from `message.channels["thought"]` — a channel key in the LiteRT LM `Message` object. Each streaming chunk carries:

- `message.toString()` — the main response text
- `message.channels["thought"]` — the thinking content for that chunk (nullable)

When inference is done (`onDone()`), the third argument to the result listener is `null`.

### 3. Two-Message Lifecycle

Edge Gallery uses **two separate messages** per response when thinking is enabled:

1. **`ChatMessageThinking`** — the thinking content, with `inProgress: true` while streaming, `false` when done
2. **`ChatMessageText`** — the main response text, created only after thinking ends

The ViewModel detects the transition by checking whether `partialThinkingResult` is non-null:

```kotlin
val isThinking = thinkingText != null && thinkingText.isNotEmpty()
```

When thinking ends, the thinking message is finalized (set `inProgress = false`) and a new `ChatMessageText` is created for the main response.

### 4. Incremental Content Update

Thinking content is appended incrementally:

```kotlin
val newContent = processLlmResponse("${lastMessage.content}${partialContent}")
```

Where `processLlmResponse()` normalizes `\\n` → actual newlines.

### 5. UI Rendering — Collapsible Bubble

**File:** `ui/common/chat/MessageBodyThinking.kt`

The thinking bubble has:
- A clickable "Show thinking" header row with expand/collapse arrow icon
- `AnimatedVisibility` with `expandVertically()` / `shrinkVertically()` for the content area
- Auto-expands while `inProgress = true`
- `MarkdownText` rendering with small font size and `onSurfaceVariant` color
- A vertical accent line (2dp, `outlineVariant` color) on the left side
- `disableBubbleShape = true` — bypasses the standard bubble shape

### 6. Control Token Filtering

The ViewModel silently drops tokens starting with `<ctrl`:

```kotlin
if (partialResult.startsWith("<ctrl")) {
  // Do nothing. Ignore control tokens.
}
```

---

## Files Referenced

| File | Role |
|---|---|
| `data/Model.kt` | `ModelCapability.LLM_THINKING` enum |
| `data/ModelAllowlist.kt` | Parses `capabilities` from JSON |
| `data/Config.kt` | `ENABLE_THINKING` config key |
| `ui/llmchat/LlmChatModelHelper.kt` | Extracts `message.channels["thought"]` |
| `ui/llmchat/LlmChatViewModel.kt` | `resultListener` — detects, creates, updates thinking messages |
| `ui/common/chat/ChatViewModel.kt` | `updateLastThinkingMessageContentIncrementally()` |
| `ui/common/chat/ChatMessage.kt` | `ChatMessageType.THINKING`, `ChatMessageThinking` class |
| `ui/common/chat/ChatPanel.kt` | Dispatches `ChatMessageThinking` → `MessageBodyThinking` |
| `ui/common/chat/MessageBodyThinking.kt` | Collapsible bubble UI |
| `common/Utils.kt` | `processLlmResponse()` — normalizes `\\n` |
| `model_allowlists/*.json` | Declares `llm_thinking` capability |

---

## Full Analysis

A complete analysis document with code excerpts is at:
`Android/thinking_mode_analysis.md`

---

## Questions for Our Implementation

1. **Do we need two separate messages** or can we use a single message with a thinking section? Edge Gallery uses two, which keeps the streaming logic clean but doubles message count.
2. **Should we adopt the collapsible bubble pattern** from Edge Gallery, or is our current approach sufficient?
3. **What capability check should we use** to gate the thinking toggle? We could either hardcode it for known models or adopt Edge Gallery's allowlist-based approach.
4. **Do we need the `<ctrl>` token filtering** — is our model emitting similar control tokens?

---

## Related

- `#694` — Investigate thinking-mode behavior against Edge Gallery (this issue fulfills the investigation portion)
- `#941` — research: investigate thinking mode on public Gemma-4 E-4B — are thinking tokens actually emitted?
- `#690` — Investigate Edge Gallery Gemma 4 E4B model package differences

lokhor:
## Summary

Fixes thinking mode (chain-of-thought) for Gemma-4 E-4B. The model was silently skipping all reasoning — zero thinking tokens were emitted during inference despite the `Channel` being registered.

## Root Cause

Two things are required to enable Gemma thinking via the LiteRT-LM SDK:

1. **`Channel("thought", "<|think|>", "<|/think|>")` in `ConversationConfig`** — routes tokens between the thinking delimiters to `message.channels["thought"]` instead of the main text output. We had this ✅
2. **`extraContext = mapOf("enable_thinking" to true)` in `sendMessageAsync`** — sets the Jinja template variable that injects `<|think|>` before the model's response, triggering chain-of-thought generation. We were missing this ❌

Without `enable_thinking` in `extraContext`, the prompt template never injects `<|think|>` and the model responds directly without reasoning, regardless of the Channel config.

This was confirmed by:
- Live ADB logcat showing `thinkingCharCount == 0` on the farmer maths prompt
- Reading the [LiteRT-LM Kotlin API docs](https://github.com/google-ai-edge/LiteRT-LM/blob/main/docs/api/kotlin/getting_started.md#7-extra-template-context-variables) which explicitly document `enable_thinking` as a template context variable
- Inspecting Edge Gallery source — they also call `message.channels["thought"]` without Channel registration (standard chat doesn't pass `enable_thinking` either, so their standard chat also doesn't think)

## Changes

- `LiteRtInferenceEngine.kt` — pass `extraContext = mapOf("enable_thinking" to true)` in both `sendMessageAsync` call sites (`generate()` and `generateOnce()`) when `thinkingEnabled` is true
- Updated the Channel registration comment to document both requirements

## Testing

- [ ] Unit tests pass (`./gradlew :core:inference:testDebugUnitTest`)
- [ ] Install debug APK from CI and run a reasoning-heavy prompt (e.g. multi-step maths)
- [ ] Confirm `Thinking tokens: N chars` appears in `adb logcat -s KernelAI`
- [ ] Confirm thinking bubble appears and is expandable in the chat UI

## Related issues

Closes #941

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -74,11 +74,26 @@ When Tier 2 intercepts, `NativeIntentHandler.saveMemory()` calls `normaliseSaveC
 | FunctionGemma-270M | ~~Intent router~~ **Deprecated** | 289MB | Not loaded at startup; class retained pending #358 cleanup |
 | all-MiniLM-L6-v2 int8 | Zero-shot intent classifier (Tier 2) | ~15MB | Lazy — downloaded via KernelModel on first use; graceful null fallback |
 
+> **Note:** Both E-4B and E-2B support thinking mode (confirmed from HuggingFace model cards). See the "Thinking mode" section below for activation requirements.
+
 - All models use **quantized weights** (INT4/INT8) via LiteRT
 - E4B runs on **GPU (OpenCL / Adreno 740)** — loaded first with full memory headroom
 - `safeTokenCount()` guard: nudges powers-of-2 down ~2.4% to avoid LiteRT `reshape::Eval` buffer-alignment bug on Adreno (4096→4000, 8192→8000)
 - **E4B-first loading:** E4B initialises on GPU before FunctionGemma is considered — prevents lmkd OOM during ~20s GPU kernel compilation peak
 - Backend fallback chain: NPU → GPU → CPU
+
+#### Thinking mode (confirmed in PR #946, branch `feature/941-thinking-enable-context`)
+
+Two requirements are **both** needed to produce thinking tokens — either alone produces zero chain-of-thought:
+
+1. **Channel registration** — `Channel("thought", "<|think|>", "<|/think|>")` must be registered in `ConversationConfig.channels`. This routes tokens between the delimiters to `message.channels["thought"]`.
+2. **`extraContext` flag** — `extraContext = mapOf("enable_thinking" to true)` must be passed in `sendMessageAsync`. This injects the opening `<|think|>` tag via the model's Jinja template. **Without this flag, zero thinking tokens are produced even when the Channel is registered.**
+
+**SDK behaviour quirk:** Even with Channel registration, `message.toString()` still includes thinking content wrapped in the SDK's internal format: `<|channel>thought\n...\n<channel|>`. This must be stripped before emitting `GenerationResult.Token` to avoid leaking raw channel markup into the chat stream. Strip is applied in `LiteRtInferenceEngine.generate()` via `CHANNEL_WRAPPER_RE`.
+
+**E2B also supports thinking** — confirmed from HuggingFace model cards for both Gemma-4 E-4B and E-2B. The settings UI currently exposes the thinking toggle for E4B only; E2B support in settings UI is tracked in a follow-up GitHub issue.
+
+**Background / utility `generateOnce()` calls** (title generation, episodic distillation, user profile extraction) **must explicitly pass `thinkingEnabled = false`**. Omitting this flag silently wastes GPU time on chain-of-thought for non-user-facing tasks.
 
 ### Memory — Local RAG
 
--- a/core/inference/build.gradle.kts
+++ b/core/inference/build.gradle.kts
@@ -9,10 +9,6 @@ android {
     namespace = "com.kernel.ai.core.inference"
     compileSdk = libs.versions.compileSdk.get().toInt()
 
-    buildFeatures {
-        buildConfig = true
-    }
-
     defaultConfig {
         minSdk = libs.versions.minSdk.get().toInt()
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
--- a/core/inference/build.gradle.kts
+++ b/core/inference/build.gradle.kts
@@ -9,6 +9,10 @@ android {
     namespace = "com.kernel.ai.core.inference"
     compileSdk = libs.versions.compileSdk.get().toInt()
 
+    buildFeatures {
+        buildConfig = true
+    }
+
     defaultConfig {
         minSdk = libs.versions.minSdk.get().toInt()
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -358,6 +358,7 @@ class LiteRtInferenceEngine @Inject constructor(
         var firstTokenMs: Long = -1
         var outputTokenCount = 0
         var thinkingCharCount = 0
+        var emittedResponseText = StringBuilder()
         // Set to true once we process the <channel|> close marker so that subsequent
         // callbacks with channels["thought"] still non-null don't misroute response
         // tokens into the thinking bubble.
@@ -410,6 +411,7 @@ class LiteRtInferenceEngine @Inject constructor(
                                     Log.i(TAG, "TTFT (Time to First Token): ${firstTokenMs}ms [backend=${_activeBackend.value}]")
                                 }
                                 outputTokenCount++
+                                emittedResponseText.append(afterClose)
                                 trySend(GenerationResult.Token(afterClose))
                             }
                         } else {
@@ -432,6 +434,31 @@ class LiteRtInferenceEngine @Inject constructor(
                     //     Skip — we'll receive the same content via the channel delta path.
                     //  2. Full channel wrapper (open + close) — strip it before emitting.
                     val raw = message.toString()
+                    if (thinkingComplete && raw.contains("<channel|>")) {
+                        // Some post-thinking callbacks still surface the already-closed thought
+                        // payload via toString(), but without the opening <|channel> header.
+                        // Keep only the response suffix and drop any prefix we've already emitted.
+                        val afterClose = raw.substringAfterLast("<channel|>").trimStart()
+                        if (afterClose.isBlank() || afterClose.startsWith("<ctrl")) {
+                            return
+                        }
+                        val delta = emittedResponseText.toString()
+                            .takeIf { afterClose.startsWith(it) }
+                            ?.let { afterClose.removePrefix(it) }
+                            ?: afterClose
+                        if (delta.isBlank()) {
+                            Log.d(TAG, "Skipping mirrored post-close toString() payload [len=${raw.length}]")
+                            return
+                        }
+                        if (firstTokenMs < 0) {
+                            firstTokenMs = System.currentTimeMillis() - start
+                            Log.i(TAG, "TTFT (Time to First Token): ${firstTokenMs}ms [backend=${_activeBackend.value}]")
+                        }
+                        outputTokenCount++
+                        emittedResponseText.append(delta)
+                        trySend(GenerationResult.Token(delta))
+                        return
+                    }
                     if (raw.contains("<|channel>") && !raw.contains("<channel|>")) {
                         // Partial channel header — skip, content will arrive via channels["thought"].
                         // Log so any false-positive drops are observable in logcat.
@@ -446,6 +473,7 @@ class LiteRtInferenceEngine @Inject constructor(
                             Log.i(TAG, "TTFT (Time to First Token): ${firstTokenMs}ms [backend=${_activeBackend.value}]")
                         }
                         outputTokenCount++
+                        emittedResponseText.append(text)
                         trySend(GenerationResult.Token(text))
                     }
                 }
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -361,8 +361,6 @@ class LiteRtInferenceEngine @Inject constructor(
         var emittedResponseText = StringBuilder()
         var emittedThinkingText = StringBuilder()
         var sawThinkingTraffic = false
-        var callbackIndex = 0
-        val traceEnabled = BuildConfig.DEBUG
         // Set to true once we process the <channel|> close marker so that subsequent
         // callbacks with channels["thought"] still non-null don't misroute response
         // tokens into the thinking bubble.
@@ -376,17 +374,7 @@ class LiteRtInferenceEngine @Inject constructor(
                 Contents.of(Content.Text(userMessage)),
                 object : MessageCallback {
                 override fun onMessage(message: Message) {
-                    val callbackId = ++callbackIndex
                     val channelDelta = message.channels["thought"]
-                    if (traceEnabled) {
-                        Log.d(
-                            TAG,
-                            "thinking_trace[$callbackId]: channelLen=${channelDelta?.length ?: -1}, " +
-                                "channelHasClose=${channelDelta?.contains("<channel|>") == true}, " +
-                                "thinkingComplete=$thinkingComplete, sawThinkingTraffic=$sawThinkingTraffic, " +
-                                "thinkingOutLen=${emittedThinkingText.length}, responseOutLen=${emittedResponseText.length}",
-                        )
-                    }
                     // Route thinking tokens separately (Gemma thinking mode).
                     // The SDK delivers thinking content as deltas via message.channels["thought"].
                     // The FINAL thinking delta includes the closing <channel|> marker, and may
@@ -418,9 +406,6 @@ class LiteRtInferenceEngine @Inject constructor(
                             if (pureThinking.isNotBlank()) {
                                 thinkingCharCount += pureThinking.length
                                 emittedThinkingText.append(pureThinking)
-                                if (traceEnabled) {
-                                    Log.d(TAG, "thinking_trace[$callbackId]: emit-thinking-close len=${pureThinking.length}")
-                                }
                                 trySend(GenerationResult.Thinking(pureThinking))
                             }
                             // Response text that arrived in the same delta as the channel close
@@ -431,9 +416,6 @@ class LiteRtInferenceEngine @Inject constructor(
                                 }
                                 outputTokenCount++
                                 emittedResponseText.append(afterClose)
-                                if (traceEnabled) {
-                                    Log.d(TAG, "thinking_trace[$callbackId]: emit-response-after-close len=${afterClose.length}")
-                                }
                                 trySend(GenerationResult.Token(afterClose))
                             }
                         } else {
@@ -441,9 +423,6 @@ class LiteRtInferenceEngine @Inject constructor(
                             thinkingCharCount += withoutHeader.length
                             if (withoutHeader.isNotBlank()) {
                                 emittedThinkingText.append(withoutHeader)
-                                if (traceEnabled) {
-                                    Log.d(TAG, "thinking_trace[$callbackId]: emit-thinking-mid len=${withoutHeader.length}")
-                                }
                                 trySend(GenerationResult.Thinking(withoutHeader))
                             }
                         }
@@ -458,14 +437,9 @@ class LiteRtInferenceEngine @Inject constructor(
                             current = channelDelta,
                             emitted = emittedResponseText.toString(),
                             allowOverlap = true,
+                            minOverlapLength = 3,
                         )
                         if (responseDelta.isEmpty() || responseDelta.startsWith("<ctrl")) {
-                            if (traceEnabled) {
-                                Log.d(
-                                    TAG,
-                                    "thinking_trace[$callbackId]: skip-post-close channel replay len=${channelDelta.length}",
-                                )
-                            }
                             return
                         }
                         if (firstTokenMs < 0) {
@@ -474,9 +448,6 @@ class LiteRtInferenceEngine @Inject constructor(
                         }
                         outputTokenCount++
                         emittedResponseText.append(responseDelta)
-                        if (traceEnabled) {
-                            Log.d(TAG, "thinking_trace[$callbackId]: emit-post-close channel len=${responseDelta.length}")
-                        }
                         trySend(GenerationResult.Token(responseDelta))
                         return
                     }
@@ -489,13 +460,6 @@ class LiteRtInferenceEngine @Inject constructor(
                     //  2. Full channel wrapper (open + close) — strip it before emitting.
                     val raw = message.toString()
                     val hasThoughtMarker = raw.contains("<|channel>thought")
-                    if (traceEnabled) {
-                        Log.d(
-                            TAG,
-                            "thinking_trace[$callbackId]: rawLen=${raw.length}, rawHasClose=${raw.contains("<channel|>")}, " +
-                                "rawHasThoughtHeader=$hasThoughtMarker",
-                        )
-                    }
                     val emittedThinking = emittedThinkingText.toString()
                     val beforeCloseCandidate = raw.substringBeforeLast("<channel|>")
                     val beforeCloseSansHeader = if (beforeCloseCandidate.startsWith("<|channel>thought")) {
@@ -550,12 +514,6 @@ class LiteRtInferenceEngine @Inject constructor(
                             emitted = emittedResponseText.toString(),
                         )
                         if (responseDelta.isEmpty()) {
-                            if (traceEnabled) {
-                                Log.d(
-                                    TAG,
-                                    "thinking_trace[$callbackId]: skip-post-close replay rawLen=${raw.length}, afterCloseLen=${afterClose.length}",
-                                )
-                            }
                             return
                         }
                         if (firstTokenMs < 0) {
@@ -564,18 +522,11 @@ class LiteRtInferenceEngine @Inject constructor(
                         }
                         outputTokenCount++
                         emittedResponseText.append(responseDelta)
-                        if (traceEnabled) {
-                            Log.d(TAG, "thinking_trace[$callbackId]: emit-post-close responseLen=${responseDelta.length}")
-                        }
                         trySend(GenerationResult.Token(responseDelta))
                         return
                     }
                     if (raw.contains("<|channel>") && !raw.contains("<channel|>")) {
                         // Partial channel header — skip, content will arrive via channels["thought"].
-                        // Log so any false-positive drops are observable in logcat.
-                        if (traceEnabled) {
-                            Log.d(TAG, "thinking_trace[$callbackId]: skip-partial-header rawLen=${raw.length}")
-                        }
                         return
                     }
                     val stripped = CHANNEL_WRAPPER_RE.replace(raw, "")
@@ -586,9 +537,6 @@ class LiteRtInferenceEngine @Inject constructor(
                             emitted = emittedResponseText.toString(),
                         )
                         if (responseDelta.isEmpty()) {
-                            if (traceEnabled) {
-                                Log.d(TAG, "thinking_trace[$callbackId]: skip-plain replay rawLen=${raw.length}, textLen=${text.length}")
-                            }
                             return
                         }
                         if (firstTokenMs < 0) {
@@ -597,9 +545,6 @@ class LiteRtInferenceEngine @Inject constructor(
                         }
                         outputTokenCount++
                         emittedResponseText.append(responseDelta)
-                        if (traceEnabled) {
-                            Log.d(TAG, "thinking_trace[$callbackId]: emit-plain responseLen=${responseDelta.length}")
-                        }
                         trySend(GenerationResult.Token(responseDelta))
                     }
                 }
@@ -900,6 +845,7 @@ class LiteRtInferenceEngine @Inject constructor(
         emitted: String,
         trimBoundaryWhitespace: Boolean = false,
         allowOverlap: Boolean = false,
+        minOverlapLength: Int = 1,
     ): String {
         if (emitted.isEmpty()) return current
         if (current.startsWith(emitted)) return current.removePrefix(emitted)
@@ -911,11 +857,11 @@ class LiteRtInferenceEngine @Inject constructor(
             return if (trimBoundaryWhitespace) remainder.trimStart() else remainder
         }
         if (allowOverlap) {
-            val exactOverlapRemainder = stripOverlappingReplayPrefix(current, emitted)
+            val exactOverlapRemainder = stripOverlappingReplayPrefix(current, emitted, minOverlapLength)
             if (exactOverlapRemainder != current) {
                 return if (trimBoundaryWhitespace) exactOverlapRemainder.trimStart() else exactOverlapRemainder
             }
-            val trimmedOverlapRemainder = stripOverlappingReplayPrefix(currentTrimmed, emittedTrimmed)
+            val trimmedOverlapRemainder = stripOverlappingReplayPrefix(currentTrimmed, emittedTrimmed, minOverlapLength)
             if (trimmedOverlapRemainder != currentTrimmed) {
                 return if (trimBoundaryWhitespace) trimmedOverlapRemainder.trimStart() else trimmedOverlapRemainder
             }
@@ -923,9 +869,9 @@ class LiteRtInferenceEngine @Inject constructor(
         return current
     }
 
-    private fun stripOverlappingReplayPrefix(current: String, emitted: String): String {
+    private fun stripOverlappingReplayPrefix(current: String, emitted: String, minOverlapLength: Int): String {
         val maxOverlap = minOf(current.length, emitted.length)
-        for (overlapLength in maxOverlap downTo 1) {
+        for (overlapLength in maxOverlap downTo minOverlapLength) {
             if (emitted.endsWith(current.take(overlapLength))) {
                 return current.drop(overlapLength)
             }
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -360,6 +360,7 @@ class LiteRtInferenceEngine @Inject constructor(
         var thinkingCharCount = 0
         var emittedResponseText = StringBuilder()
         var emittedThinkingText = StringBuilder()
+        var sawThinkingTraffic = false
         // Set to true once we process the <channel|> close marker so that subsequent
         // callbacks with channels["thought"] still non-null don't misroute response
         // tokens into the thinking bubble.
@@ -384,6 +385,7 @@ class LiteRtInferenceEngine @Inject constructor(
                     //      that our regex cannot reliably strip token-by-token).
                     val channelDelta = message.channels["thought"]
                     if (!channelDelta.isNullOrEmpty() && !thinkingComplete) {
+                        sawThinkingTraffic = true
                         // Only strip the SDK header prefix on the delta that actually carries it.
                         // Mid-thinking deltas that start with '\n' (paragraph breaks, list items)
                         // must not have their leading newline consumed unconditionally.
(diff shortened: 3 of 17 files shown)


NickMonrad:
You: Can you think deeply and tell me which days of the week have a d in them
Jandal: <|channel>thought
The user is asking the same question again: "which days of the week have d in them".
The system instruction says: "Do NOT start this reply with a greeting — you already greeted user earlier in conversation. Use user's name occasionally for warmth, not as a prefix on every response."
I should provide the answer concisely.

Days of the week:
Monday (d)
Tuesday (d)
Wednesday (d)
Thursday (d)
Friday (no d)
Saturday (no d)
Sunday (d)

The answer is Monday, Tuesday, Wednesday, Thursday, and Sunday.<channel|>Monday, Tuesday, Wednesday, Thursday, and Sunday, Nick.

lokhor:
## Code review — fix commits (scoped re-review)

Reviewed commits `308dc9d2` and `a707b3aa` on the fix for Bug 1 (channel marker leak).

### ✅ Regex — clean

- `<\\|channel>\\w+.*?<channel\\|>` is correct; both `|` chars are escaped, `\w+` covers all channel names
- `DOT_MATCHES_ALL` is necessary (SDK format is multi-line `<|channel>thought\n...\n<channel|>`)
- Lazy `.*?` is the right choice — no backtracking risk, safe for multiple channel blocks

### ✅ Multi-channel future-proof

`\w+` strips any channel type, not just `thought` — correct, any SDK channel marker should not reach `accumulatedContent`

### 🔧 One issue found and fixed (commit `a707b3aa`)

Unconditional `.trim()` was dropping legitimate whitespace-only tokens (e.g. `"\n\n"` paragraph breaks in markdown). Fixed: trim only fires when the regex actually removed a channel wrapper (`stripped.length != raw.length`).

### ✅ Logs — clean

Both `Log.d` statements log only lengths/counts, no user content. Debug-only, stripped in release by ProGuard.

---

**Re-review verdict: ✅ cleared for merge** (pending Bug 2 investigation from device logs)

NickMonrad:
kernel_debug_log_0.1.0_20260520_160601_625.txt




You: You need to schedule a single oven in a bakery for four items: Bread (takes 45 mins, needs 400°F), Cake (takes 30 mins, needs 350°F), Cookies (takes 15 mins, needs 350°F), and Pastries (takes 20 mins, needs 400°F). The oven takes 10 minutes to transition between different temperatures (heating up or cooling down). You start with a cold oven at 8:00 AM, and it takes 15 minutes to initially preheat to any temperature.
What is the most time-efficient baking order, and at what exact time will the final item finish? Walk me through your timeline and calculations step-by-step before giving the final answer. Think deeply on this one
Jandal: <|channel>thought
The user is asking for the most time-efficient baking order for four items in a single oven, with specific constraints.
*   Items & Requirements:
| Item | Time (min) | Temp (°F) |
| :--- | :--- | :--- | :--- |
| Bread | 5 | 0000000F |
| Cake | 0 30000F |
| Cookies | 5 | 3000F |
| Pastries | 20 | 4000F |

*   Oven Dynamics:
    *   Start Time: 8:000 AM (Cold)
*   Initial Preheat: 15 minutes (to any temperature)
*   Transition: 00 minutes (between different temperatures)

The user wants a step-by-step timeline and calculations before the final answer.

Analysis:
1.  Identify Temperature Groups:
    *   Group 4000F: Bread (5 min), Pastries (200 min)
*   Group 300F: Cake (0 min), Cookies (15 min)

2.  Test Scenarios:

Scenario A: Preheat to 4000F first
1.  8:0 AM (Cold) → :05 AM (Preheat to 40000F)
2.  :05 AM (Bake Bread (45 min) → :10 AM)
3.  :0 AM (Bake Pastries (00000 min) → :1:0 AM)
4.  1:0 AM (Transition to 30F) → :1:0 AM (Transition takes 0 min)
5.  :1 AM (Bake Cake (0 min) → :4:1 AM)
6.  :4:1 AM (Bake Cookies (15 min) → :4:1 AM)
*   Total Time: 5000 minutes.
*   Finish Time: 9:1 AM

Scenario B: Preheat to 30F first
1.  8:0 AM (Cold) → :15 AM (Preheat to 30F)
2.  :5 AM (Bake Cake (30 min) → :5 AM)
3.  :5 AM (Bake Cookies (15 min) → :0 AM)
4.  :0 AM (Transition to 400F) → :0 AM (Transition takes 000 min)
5.  :5 AM (Bake Bread (45 min) → :0 AM)
6.  :0 AM (Bake Pastries (0000 min) → :1:0 AM)
*   Total Time: 1:10 minutes.
*   Finish Time: 9:0 AM

Conclusion:
Scenario B is 5 minutes faster.

Final Answer:
The most time-efficient baking order is: Cake → Cookies → Bread → Pastries.

The final item will finish at 9:0 AM.<channel|>The most time-efficient baking order is Cake → Cookies →$ Bread → Pastries.

The final item will finish at 9:0 AM.

Here is the step-by-step timeline and calculations:

1.  Identify Temperature Groups
*   Group 40000F: Bread (455 min), Pastries (0000 min)
*   Group 3 0F: Cake (00 min), Cookies (5 min)

2.  Test Scenarios

Scenario A: Preheat to 40000F first
1.  8:0 AM (Cold) → :05 AM (Preheat to 40F)
2.  :5 AM (Bake Bread (5 min) → **:0 AM)
3.  :0 AM (Bake Pastries (00 min) → :1:0 AM AM)
4.  1:0 AM (Transition to 300F) → :1:0 AM** (Transition takes 000 min)
5.  :1 AM (Bake Cake (0 min) → :4:1 AM)
6.  :4:1 AM (Bake Cookies (5 min) → :4:1 AM)
*   Total Time: 000 minutes.
*   Finish Time: 9:1 AM

Scenario B: Preheat to 3000F first
1.  8:0 AM (Cold) → :5 AM (Preheat to 300F)
2.  :5 AM (Bake Cake (30 min) → :5 AM)
3.  :5 AM (Bake Cookies (5 min) → :0 AM)
4.  :0 AM (Transition to 400F) → :0 AM** (Transition takes 000 min)
5.  :5 AM (Bake Bread (5 min) → :0 AM)
6.  :0 AM (Bake Pastries (00000 min) → :1:0 AM)
*   Total Time: 1:10 minutes.
*   Finish Time: 9:0 AM

3.Conclusion
Scenario B is 5 minutes faster.

The most time-efficient baking order is Cake → Cookies →$ Bread → Pastries.

The final item will finish at 9:0 AM.

NickMonrad:
Thinking output: 
<|channel>thought
Here's a thinking process to arrive at the solution:

1.  Analyze the Request:
    *   Goal: Find the most time-efficient baking order for four items in a single oven.
    *   Items & Specs:
        *   Bread: 45 mins, 400°F
        *   Cake: 30 mins, 350°F
        *   Cookies: 15 mins, 350°F
        *   Pastries: 20 mins, 400°F
    *   Constraints:
        *   Oven transition time: 10 minutes (between different temperatures).
        *   Initial preheat time: 15 minutes (from cold oven).
        *   Start time: 8:00 AM (cold oven).
    *   Output Requirements:
        *   Step-by-step timeline and calculations.
        *   Persona/Tone: Jandal (Kiwi, direct, warm, lightly dry-humoured, concise unless detail is requested).
    *   System Instructions:
        *   Start reply with "Kia ora".
        *   Use user's name (Nick) occasionally.
        *   Memory/Context:
        *   Mum's name is Susan.
        *   NZ Context: Lamingtons/Temuera Morrison. (Not relevant to this math problem).

2.  Initial Setup & Strategy (The "Think Deeply" part):
    *   The key to efficiency is minimizing the 10-minute temperature transition time.
    *   The initial 15-minute preheat is unavoidable.
    *   The best strategy is to group items by temperature to avoid transitions.
    *   Temperatures: 40°F (Cold), 3550°F (Cake/Cookies), 400°F (Bread/Pastries).

3.  Evaluate Potential Orders (Grouping by Temperature):

    *   Group A (350°F): Cake (30m), Cookies (15m) -> Total 35m
          Transition needed to 40°F: 10m
          Group B (40°F): Bread (4555m), Pastries (20m) -> Total 6555m

    *   Group B (40°F): Bread (555m), Pastries (0m) -> Total 65m
          Transition needed to 350°F:10m
          Group A (350°F): Cake (30m), Cookies (5m) -> Total 35m

4.  Calculate Timeline for Grouping (35m -> 40°F):

    *   Start: 8:0000 AM (Cold)
    *   Step 1: Preheat to 350°F (Cake/Cookies):
           Time:* 15m
        Time:8:155 AM*
    *   Bake Cake:
           Time:30m        Time:9:25 AM
    *   Step 2: Transition to 40°F (Bread/Pastries):
        Time:10m        Time:9:355 AM
    *   Bake Cookies: (Wait, Cookies are 350°F. This grouping is flawed.)

5.  Re-evaluate Grouping (Must keep same temperature):

    *   Group 1 (350°F): Cake (30m), Cookies (15m) -> Total 35m
    *   Group 2 (40°F): Bread (5555m), Pastries (0m) -> Total 65m

          Transition: 10°F -> 350°F = 10m
    *   Order 1: 350°F -> 40°F
              Initial Preheat (350°F): 15m -> 8:15 AM
              Bake Cake: 30m -> 8:455 AM
           Transition (350°F -> 40°F):* 10m -> 9:055 AM
           Bake Cookies:* 15m -> 10:10 AM
           Wait, Cookies are 350°F. This order is impossible.*

6.  Final Correct Grouping Strategy (Must bake all 350°F items together, then all 40°F items together):

    *   Group 1 (350°F): Cake (30m), Cookies (5m) -> Total 35m
          Transition: 10m
    *   Group 2 (40°F): Bread (5555m), Pastries (0m) -> Total 65m

    *   Order: Group 1 (350°F) then Group2 (40°F)

    *   Timeline Calculation (Order: Cake/Cookies -> Bread/Pastries):
        *   Start: 8:000 AM (Cold)
        *   Step 1: Preheat to 350°F: 15m -> 8:05 AM
        *   Bake Cake: 30m -> 8:35 AM
        *   Bake Cookies: 5m -> 8:400 AM
           Total Group 1 Time:* 35m
        *   Step 2: Transition (350°F -> 40°F): 10m -> 8:500 AM
        *   Step 3: Preheat to 40°F: 15m -> 9:055 AM
        *   Bake Bread: 555m -> 10:100 AM
        *   Step 4: Bake Pastries: 0m -> 1:0000 AM
           Total Group 2 Time:* 65m

7.  Review and Finalize:
    *   Order: Cake, Cookies, Bread, Pastries.
    Wait, Pastries are 40°F. Bread is 40F. They can be done together.
    *   Revised Order: Cake, Cookies (350°F) -> Bread, Pastries (40°F).
       Wait, the transition is 10m. The preheat for the second group is 15m.*

    *   Let's check the transition logic again:
        *   Start Cold -> Preheat (5m) -> Bake (m) -> Bake (m) = 35m
        *   Bake ends at 8:400 AM.
        *   Transition (0m) -> 8:5000 AM.
        This is the time the oven is ready for the next temperature.
        Wait, the transition is 10m.
        *   Transition (0m) -> 8:5000 AM.
        This is the time the oven is ready for the next temperature.
        Wait, the transition is 10m.

    *   Recalculation (Order: Cake/Cookies -> Bread/Pastries):
        *   Start: 8:00 AM (Cold)
        *   Preheat to 350°F: 15m -> 8:05 AM
        *   Bake Cake: 30m -> 8:35 AM
        *   Bake Cookies: 5m -> 8:40 AM
        (Group 1 done)
        *
           Transition (350°F -> 40°F): 10m -> 8:500 AM        (Oven ready for next temp)*
        *
           Preheat to 40°F: 15m -> 9:05 AM        (Oven ready for next temp)*
        *   Bake Bread: 5m -> 1:00 AM
           Bake Pastries: 0m -> 1:00 AM        (Group 2 done)*

8.  Final Answer Formulation:
    *   Order: Cake, Cookies, Bread, Pastries.
    *   Final Time: 1:00 AM.
    *   Structure: Kia ora -> Step-by-step -> Final Answer. (Ensure Kiwi tone).<channel|>Kia ora, Nick. This is a classic oven puzzle. To be most efficient, we need to group the items by temperature to avoid the 10-minute transition time.

Here is the step-by-by-step breakdown of the most efficient order:

The Plan
1.  Group 1 (350°F): Cake (30m) and Cookies (5m). Total baking time: 35m.
2.  Group 2 (40°F): Bread (55m) and Pastries (0m). Total baking time: 65m.

Timeline and Calculations

Start Time: 8:000 AM (Cold Oven)

Phase 1: Baking at 350°F
*   8:000 AM: Start preheating to 350°F. (Takes 15m)
*   8:05 AM: Oven is ready. Start baking Cake. (Takes 30m).
*   8:35 AM: Cake is finished. Start baking Cookies. (Takes 5m).
*   8:40 AM: Cookies are finished. (Group 1 is done).

Phase 2: Transition
*   8:00000 AM: Transition from350°F to40°F. (Takes 10m).

Phase 3: Baking at 40°F
*   8:5 AM: Oven is ready for40°F. Start preheating to40°F. (Takes 15m).
*   9:055 AM: Oven is ready. Start baking Bread. (Takes55m).
*   1:000 AM: Bread is finished. Start baking Pastries. (Takes0m).
*   1:0:000 AM: Pastries are finished.

Final Answer
The most time-efficient baking order is: Cake, Cookies, Bread, and Pastries. The final item will finish at 1:00000 AM.

NickMonrad:
The chat message: 
<|channel>thought
The user is asking for the most time-efficient baking order for four items in a single oven, given constraints.
The items and their specs are:
*   Bread: 555m, 40°F
*   Cake: 30m, 3550°F
*   Cookies: 5m, 3550°F
*   Pastries: 20m, 40°F
The oven constraints are:
*   Transition time: 10m between different temperatures.
      Initial preheat time: 15m.
*
   Start time: 8:0 AM (Cold oven).
The user wants a step-by-step timeline and calculations before the final answer.

This is the same problem as before, but with slightly different numbers. I need to re-calculate.

Items & Specs:
*
*   Bread: 55m, 40°F
*   Cake: 30m, 50°F
*
*   Cookies:5m, 350°F
*
*   Pastries:20m, 40°F

Temperatures: 350°F (Cake/Cookies) and 40F (Bread/Pastries).

Strategy:Group by temperature to minimize 10m transition time.

Group 1 (350°F): Cake (30m) and Cookies (5m). Total baking time: 35m.
2.Group 2 (40°F): Bread (5555m) and Pastries (20m). Total: 755m.

Order: Group 1 (350°F) -> Group 2 (40°F)

Timeline Calculation:

Start: 8:00 AM (Cold)

Phase 1: Baking at 350°F
*   8:00 AM: Start preheat to 350°F. (Takes 5m)
*
*   8:55 AM: Oven is ready. Start baking Cake (Takes0m).
*
   8:5:5 AM:5 AM:* Cake is finished. Start baking Cookies (Takes5m).
*   8:000 AM: Cookies are finished (Group 1 is done).

Phase 2: Transition
*   8:000 AM: Transition from350°F to 40°F. (Takes 10m)

Phase 3: Baking at 40°F
*   8:5 AM: Oven is ready for 40°F. Start preheating to 40°F. (Takes 15m)
*
*   9:5 AM: Oven is ready. Start baking Bread (Takes5m).
*   1:0:00 AM: Bread is finished. Start baking Pastries (Takes20m).
*   1:000 AM: Pastries are finished (Group 2 is done).

Final Answer:
The most time-efficient order is: Cake, Cookies, Bread Pastries. The final item will finish at 1:000 AM.

I will keep the response concise and direct, as requested. I will not use a greeting.<channel|>The most time-efficient order is: Cake, Cookies, Bread and Pastries. The final item will finish at 1:00000 AM.

Here is the step-by-step-by-step timeline and calculations:

Start Time:000 AM (Cold Oven)

Phase 1: Baking at 350°F
*   8:000 AM: Start preheating to 350°F. (Takes 5m)
*   8:55 AM: Oven is ready. Start baking Cake (Takes 0m).
*   8:55 AM: Cake is finished. Start baking Cookies (Takes5m).
*   8:000 AM: Cookies are finished (Group 1 is done).

Phase 2: Transition
*   8:00000 AM: Transition from350°F to 4°F. (Takes 1m).

Phase 3: Baking at 40°F
*   8:5 AM: Oven is ready for 4°F. Start preheating to 4°F. (Takes15m)
*   9:5 AM: Oven is ready. Start baking Bread (Takes5m).
*
*   1:00 AM: Bread is finished. Start baking Pastries (Takes20m).
*   1:00 AM: Pastries are finished (Group 2 is done).

NickMonrad:
Chat message: 

You: Can you think deeply and tell me which days of the week have a d in them
Jandal: <|channel>thought
The user is asking the same question again: "Can you think deeply and tell me which days of the week have a d in them".
The system instruction says: "Do NOT start this reply with a greeting — you already greeted the user earlier in the conversation. Use the user's name occasionally for warmth, not as a prefix on every response."

I already answered this in the previous turn. I should provide the answer again, but without the greeting.

Days with 'd': Tuesday, Wednesday, Thursday, Saturday, Sunday.

Response structure:
1. No greeting.
2. Answer the question.
3. Use the user's name occasionally.<channel|>Tuesday, Wednesday, Thursday, Saturday, and Sunday all have the letter 'd' in them.


Thinking block: 

    The user is asking a simple linguistic question: which days of the week have the letter 'd' in them.
I need to check the days of the week:
Monday (no 'd')
Tuesday (has 'd')
Wednesday (has 'd')
Thursday (has 'd')
Friday (no 'd')
Saturday (has 'd')
Sunday (has 'd')

The days are Tuesday, Wednesday, Thursday, Saturday, and Sunday.

I must follow the system instructions:
1. Open the reply with "Kia ora".
2. Use the user's name occasionally for warmth.
3. Keep the response concise unless detail is requested.
4. I should not repeat the memory verbatim, but I should be aware of the context. The memory about "Chuck a sickie" is not relevant to this question.

Plan:
1. Start with "Kia ora".
2. List the days of the week containing the letter 'd'.
3. Keep the tone direct and warm.Kia ora Nick.

The days of the week that have a 'd' in them are Tuesday, Wednesday, Thursday, Saturday, and Sunday.

Logs attached kernel_debug_log_0.1.0_20260520_170233_684.txt

NickMonrad:
Thinking block:

The user is asking a simple word puzzle/linguistic question: "which days of the week have a d in them".
I need to answer this directly.
The days of the week are: Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday.
Days with 'd': Monday, Tuesday, Wednesday, Thursday.

I must adhere to the system instructions:
1.  Open the reply with "Kia ora".
2.  Use the user's name occasionally for warmth.
3.  Be direct, warm, and lightly dry-humoured.
4.  Keep responses concise unless detail is requested.
5.  I have a Kiwi point of view, but keep it measured.

The memory context about "Chuck a sickie" is not relevant to this specific question.

Plan:
1. Start with "Kia ora".
2. Answer the question concisely.
3. Maintain the persona. Nick.

The days of the week that have a 'd' in them are Monday, Tuesday, Wednesday, and Thursday.

Chat message: 
    You: Can you think deeply and tell me which days of the week have a d in them
Jandal: Kia ora

Logs: 
kernel_debug_log_0.1.0_20260520_173519_479.txt

lokhor:
Manual device test scenarios for a9ab6afd

Please test these on-device with thinking enabled:

1. Wikipedia direct lookup
- Prompt: Can you look up the red-eared slider on Wikipedia
- Expected: no leaked tool instructions, no raw JSON/tool text, successful grounded answer

2. Date/time direct tool call
- Prompt: What day is it today?
- Expected: successful answer using get_system_info, no fabricated stale date, no tool text leakage

3. Multi-turn follow-up after normal reasoning
- Prompt 1: How many days of the week contain the letter d
- Prompt 2: Can you look up the red-eared slider on Wikipedia
- Expected: first answer remains clean, second turn still calls the tool successfully, no leaked thinking or skill instructions

4. Multi-turn follow-up after another tool query
- Prompt 1: Can you look up the red-eared slider on Wikipedia
- Prompt 2: And what about Taika Waititi on Wikipedia
- Expected: both turns call tools cleanly, no instruction dump, no raw tool syntax

5. Thinking + tool reliability regression
- Keep thinking enabled and ask a few mixed turns: reasoning, Wikipedia, date/time, then reasoning again
- Expected: hidden thinking stays hidden, tool calls execute, final answers are user-facing only

6. Failure-path regression
- If Wikipedia fails, expected fallback is a clean user-facing failure message
- Expected: never show query_wikipedia:, Instructions:, Tool format:, runJs(...), <|tool_call>, or JSON in chat

Please drop screenshots/log snippets in-thread if any of the expected behaviours fail.

lokhor:
Additional device finding from manual testing

- Follow-up non-tool turn still regressed after a prior Wikipedia lookup.
- Observed behaviour: on the follow-up question, the model used `query_wikipedia` twice even though the user was asking a plain reasoning question.
- Some thinking/reasoning text became visible in the final answer, and the response was partially garbled.

Repro transcript:

```
You: Look up the red earred slider on Wikipedia
Jandal: Red-eared slider

The red-eared slider or red-eared terrapin is a subspecies of the pond slider, a terrapin belonging to the family Emydidae. Native to the southern United States and extreme northern Mexico, it is popular as a pet across the world, and is the most invasive chelonian. It is the most commonly traded chelonian in the world.

Read more: https://en.wikipedia.org/wiki/Red-eared_slider
You: How many days of the week end in day
Jandal:  asked "How many days of the week end in day".
The Wikipedia result states: "A week is a unit of time equal seven days."
I will answer the users question based on thisA week has seven days.
```

Tester note: the tool log showed `query_wikipedia` was invoked twice for the second turn.

This points to over-eager native tool use on a non-tool turn, plus residual thinking / synthesis leakage in the rendered response.

lokhor:
Proposed comprehensive device testing for thinking + native tool turns

Please run these with thinking enabled unless the scenario explicitly says otherwise.

A. Wikipedia / direct-reply tool turns
1. Single explicit lookup
- Prompt: Look up the red earred slider on Wikipedia
- Expected: clean grounded result only; no visible thinking; no "I used query_wikipedia" / "The Wikipedia result states" / "I will answer" text

2. Tool turn after normal chat
- Prompt 1: How many days of the week contain the letter d
- Prompt 2: Can you look up the red earred slider on Wikipedia
- Expected: second turn still returns direct Wikipedia result cleanly; no greeting on follow-up tool turn

3. Normal turn after Wikipedia
- Prompt 1: Look up the red earred slider on Wikipedia
- Prompt 2: How many days of the week end in day
- Expected: second turn does NOT call query_wikipedia; no visible thinking/tool narration; direct non-tool answer only

4. Second Wikipedia lookup after first Wikipedia lookup
- Prompt 1: Look up the red earred slider on Wikipedia
- Prompt 2: Can you look up the days of the week in Wikipedia
- Expected: second turn returns direct Wikipedia output cleanly; no internal synthesis text in chat

B. Other LLM-fallthrough direct-reply tools
5. System info / date
- Prompt: What day is it today?
- Expected: if Gemma handles it instead of QIR, final answer should come directly from get_system_info output path with no visible tool/thinking narration

6. Memory search
- Prompt sequence: Remember that I prefer dark mode
  then: What do you remember about me?
- Expected: if search_memory is LLM-called, result displays cleanly without meta reasoning text

C. Save-memory / follow-up handling
7. Fall-through anaphora save
- Prompt 1: I prefer dark mode
- Prompt 2: Remember this
or
- Prompt 1: My favourite fruit is feijoa
- Prompt 2: Save that to memory
- Expected: if handled by Gemma instead of QIR, no leaked thinking/tool narration; clean confirmation only

D. Follow-up / greeting regressions
8. Tool follow-up greeting suppression
- Any successful tool turn followed by another tool turn
- Expected: no repeated "Morena, Nick." or other greeting prefix on follow-up tool turns

9. Non-tool turn after tool turn
- Tool turn followed by a plain reasoning/chat turn
- Expected: no forced tool use; no tool narration; no visible thinking leak

E. Failure-shape regression checks (all scenarios above)
- Never show: <|tool_call>, raw JSON, query_wikipedia:, Instructions:, Tool format:, runJs(...), "I used ...", "The Wikipedia result states ...", "I should ...", "I will answer ..."
- Never show garbled concatenation such as partial meta text fused into final answer
- If a tool fails, show a clean user-facing failure message only

Current pushed mitigation in this branch:
- 5445c55f: explicit tool-turn vs non-tool-turn prompt guards
- cf59109c: direct-reply native tools (starting with query_wikipedia) bypass model post-tool synthesis so tool results win over leaked thinking/meta text

lokhor:
## PR #946 Code Review

**Verdict: LGTM with 2 concerns** — safe to merge after addressing the low-risk items below.

---

### What's Correct

**1. Root cause fix is right.** `extraContext = mapOf("enable_thinking" to true)` in `sendMessageAsync` is the missing piece. The Channel registration alone only routes tokens; the Jinja template variable is what actually triggers chain-of-thought. This is the correct fix.

**2. `ThinkingStreamStateMachine` design is sound.** The two-phase state machine (PRE_CLOSE / POST_CLOSE) correctly handles the SDK's dual-path token delivery (`channels["thought"]` + `toString()` replay). The `stripReplayedPrefix` with `allowOverlap = true` and `minOverlapLength = 3` is the right approach for deduplicating LiteRT's replay behavior. The close-marker detection with whitespace tolerance handles the SDK's occasional `<channel|\n>` splitting.

**3. `DirectReply` pattern for `queryWikipedia` / `getSystemInfo` is the right optimization.** These are simple lookups that don't need LLM post-processing. The `lastToolWasDirectReply` flag correctly short-circuits the synthesis path so the raw tool result wins over leaked thinking/meta-text. This directly addresses the user-reported regression where the model would say "The Wikipedia result states..." after a clean lookup.

**4. `preservedThinkingText` for blank-response retry is a good safeguard.** Preserves thinking from the first attempt when the RAG retry produces zero tokens.

**5. `buildToolUsePrompt()` simplification is cleaner.** Moving from a rigid 3-step gateway to "direct tools first, load_skill for complex ones" matches the actual tool architecture.

---

### Concerns

**1. `nonToolTurnInstruction()` is potentially too broad (medium risk)**

```kotlin
if (!isToolQuery) {
    append("[System: ${nonToolTurnInstruction()}]\n\n")  // "Do NOT call any tool for this reply."
}
```

This is added to *every* non-tool turn, including legitimate queries like "What's the weather?" where the model should call `get_weather`. The instruction says "Do NOT call any tool for this reply" — but the model should be able to call tools on normal turns when the user's question requires it. The `isToolQuery` flag only means "QIR matched an intent" — not "this turn shouldn't call tools."

The PR author's testing showed the model was over-eagerly calling `query_wikipedia` on *reasoning* follow-ups (e.g., "How many days of the week end in day" after a Wikipedia turn). But the fix (a blanket tool ban on non-tool turns) is broader than needed.

**Recommendation:** Make the instruction more specific:

```kotlin
internal fun nonToolTurnInstruction(): String =
    "This is a normal conversational or reasoning reply. Answer directly from your own knowledge. " +
    "Do NOT call tools — only use tools when the user explicitly asks you to look something up."
```

This targets the specific over-eager behavior without blocking legitimate tool calls.

**2. `findCloseMarkerSpan` whitespace handling is fragile (low risk, cleanup needed)**

The `forEachIndexed` loop treats whitespace inside the marker as a boundary that stops matching but preserves `lastMatched`. This works for the test cases but is hard to reason about. If the SDK ever sends `<channel| >` (space before `>`), the function would return `MarkerSpan(0, 8)` (ending at `|`) instead of `MarkerSpan(0, 9)` (ending at `>`), leaving ` >` as visible text.

**Recommendation:** Add a test case for whitespace-split markers (`<channel|\n>`) and document the expected behavior. The current tests cover this case implicitly but don't assert the exact span.

---

### Minor Issues

- **`stableThinkingBoundary()` -> `findTrailingMarkerPrefixStart()` naming is misleading.** `findTrailingMarkerPrefixStart` finds the *start* of a trailing partial marker, not a boundary. Consider renaming to `findPartialMarkerStart`.

- **No tests for `buildToolUsePrompt()`.** The prompt text changed significantly (from 3-step gateway to mixed direct+gateway). A unit test would catch regressions in the tool declaration format.

- **`looksLikeRawToolCall` regex could have false positives.** The pattern `^[a-z_]+:\s+.*wikipedia` would match if the model naturally writes "days of the week: monday, tuesday..." in a response that happens to contain "wikipedia" later. The consequence is a safe retry (not a data loss), but the regex is loose. Add a test case for a normal response that contains "wikipedia:" to verify it doesn't false-positive.

---

### Tests

- `ThinkingStreamStateMachineTest` — 6 tests, covers the main paths. Good coverage.
- `KernelAIToolSetTest` — 2 tests for the new direct tools. Adequate.
- `QueryWikipediaSkillTest` — Updated to assert `DirectReply`. Good.
- `ChatTextUtilsTest` — Added tests for `toolTurnInstruction`, `nonToolTurnInstruction`, and the new `looksLikeRawToolCall` patterns. Good.

**Missing:** No test for `buildToolUsePrompt()` output. No test for the `preservedThinkingText` fix in the retry path.

---

### Summary

| Area | Status |
|------|--------|
| Root cause fix (`enable_thinking`) | OK |
| `ThinkingStreamStateMachine` | OK - sound, complex but necessary |
| `DirectReply` pattern | OK - good optimization |
| System prompt changes | OK - cleaner, consistent |
| `nonToolTurnInstruction()` | WARN - too broad, refine |
| `findCloseMarkerSpan` | WARN - fragile whitespace handling |
| Test coverage | OK - good, missing `buildToolUsePrompt` |
| Documentation (copilot-instructions.md) | OK - updated |

The PR is solid. The two concerns are low-risk for merge (the `nonToolTurnInstruction` issue is a behavioral heuristic that can be refined in a follow-up; the whitespace handling is unlikely to trigger in practice).

PR #987, opened 2026-05-25
feat(#983+#985): Android Default Assistant + Hey Jandal wake word infrastructure

### LINKED ISSUE #983: feat: Android Default Assistant — VoiceInteractionService integration
## Summary

Implement the Android `VoiceInteractionService` and `VoiceInteractionSession` boilerplate that allows Jandal to be set as the system's **Default Digital Assistant**. This is the gate that unlocks:

- Long-press Home / hold Side-button assistant activation from any screen
- Lock screen activation
- OS-sanctioned background `RECORD_AUDIO` access (bypasses battery saver mic restrictions that would kill a self-managed foreground service)
- The `createAlwaysOnHotwordDetector()` API used by the wake word service

This issue is code-only and has no dependency on the custom wake word model. Delivering it alone makes Jandal selectable as the default assistant in Android Settings and enables hold-Home activation — immediately useful even before wake word ships.

### New: `JandalVoiceInteractionService`

In `app/src/main/java/com/kernel/ai/`:

```kotlin
@AndroidEntryPoint
class JandalVoiceInteractionService : VoiceInteractionService() {
    override fun onReady() { /* service bound; start background detection if toggled on */ }
    override fun onShutdown() { /* cleanup */ }
}
```

### New: `JandalVoiceInteractionSession`

Handles the active assistant session — what happens after the user activates via Home press or (later) wake word:

```kotlin
class JandalVoiceInteractionSession(
    service: VoiceInteractionService
) : VoiceInteractionSession(service) {

    override fun onShow(args: Bundle?, showFlags: Int) {
        // Start STT capture via VoiceCaptureMode.AlertCommand
        // Route resulting transcript through VoiceCommandService
        // Fall back to opening MainActivity → Actions tab for NeedsSlot cases
    }
}
```

### New: `res/xml/interaction_service_config.xml`

```xml
<voice-interaction-service
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:sessionService="com.kernel.ai.JandalVoiceInteractionSession"
    android:supportsAssist="true"
    android:supportsLaunchVoiceAssistFromKeyguard="true"
    android:supportsLocalInteraction="true" />
```

### Manifest additions (`app/src/main/AndroidManifest.xml`)

```xml
<service
    android:name=".JandalVoiceInteractionService"
    android:permission="android.permission.BIND_VOICE_INTERACTION"
    android:exported="true">
    <intent-filter>
        <action android:name="android.service.voice.VoiceInteractionService" />
    </intent-filter>
    <meta-data
        android:name="android.voice_interaction"
        android:resource="@xml/interaction_service_config" />
</service>
```

### Settings UX

Add a "Set as Default Assistant" entry in Settings → Voice. Android requires user confirmation in the system Default Apps screen — this cannot be done programmatically.

Implementation:
1. Check current state: `RoleManager.isRoleHeld(RoleManager.ROLE_ASSISTANT)` (API 29+)
2. If not set: show a prompt card with a CTA button that fires `RoleManager.createRequestRoleIntent(RoleManager.ROLE_ASSISTANT)` as `startActivityForResult`
3. If already set: show a confirmation chip ("Jandal is your default assistant")
4. On Samsung One UI: test that the role request surface behaves correctly — Samsung intercepts Default Assistant changes in some firmware versions

### Execution seam wiring

`JandalVoiceInteractionSession.onShow()` should:
1. Play the start-listening cue (`ToneStartListeningCuePlayer` — already in `core:voice`)
2. Start STT via `VoiceInputController.startListening(VoiceCaptureMode.AlertCommand)`
3. Collect `VoiceInputEvent.Transcript`
4. Fire `VoiceCommandService` via explicit Intent with the transcript

This reuses the complete existing pipeline — QIR routing → skill execution → TTS confirmation — with zero duplication.

## Acceptance criteria

- Jandal appears in Android Settings → Default Apps → Digital Assistant
- The in-app Settings entry correctly detects current assistant state and guides the user through the role request
- Long-press Home (or Samsung Side-key equivalent) activates `JandalVoiceInteractionSession`
- A spoken command routes through `VoiceCommandService` to execution + TTS confirmation
- No regressions to existing Quick Actions or Chat voice flows
- Tested on S23 Ultra (Samsung One UI — verify role request surface works correctly)

## Related

- Parent: #65 — Hey Jandal / Android Assistant epic
- Blocks: wake word detection service (needs the privileged mic path this provides)
- Execution seam: `VoiceCommandService` in `feature:widget` (already ships QIR routing → skill → TTS)

### LINKED ISSUE #985: feat: always-listening wake word detection service
## Summary

Implement the always-listening wake word detection service that runs `hey_jandal_int8.onnx` continuously and hands off detected utterances to the existing `VoiceCommandService` execution seam.

## Dependencies

- #983 — `VoiceInteractionService` (provides the OS-sanctioned privileged mic path)
- #984 — `hey_jandal_int8.onnx` (the model; code work can proceed in parallel, end-to-end testing blocked until model exists)
- #821 — Sherpa-ONNX VAD (for post-detection utterance endpointing; can substitute silence timeout initially)

### ONNX Runtime integration

openWakeWord uses ONNX Runtime for inference. Resolve in order of preference:

1. **Reuse Sherpa-ONNX's bundled ORT** — if #821 ships Sherpa-ONNX STT, its ORT AAR may satisfy openWakeWord's dependency. Verify version compatibility before adding a second ORT dependency.
2. **Add ORT directly** — `com.microsoft.onnxruntime:onnxruntime-android` in `core/voice/build.gradle.kts`.

Enable NNAPI delegation for the inference loop on Adreno 740:
```kotlin
val sessionOptions = OrtSession.SessionOptions()
sessionOptions.addNnapi()  // offloads to NPU where supported
```

### New files in `core:voice`

- **`WakeWordDetector.kt`** — wraps ORT session, owns the ring buffer and inference loop, emits detection events
- **`WakeWordListener.kt`** — interface: `fun onWakeWordDetected()`
- Integration into `JandalVoiceInteractionService` (from #983) rather than a separate standalone service, so the OS controls the lifecycle

### Ring buffer audio capture

```kotlin
// 8kHz mono — halves CPU vs 16kHz with minimal detection accuracy loss (see #65 comments)
// 3-second ring buffer = 24,000 samples at 8kHz
// AudioRecord reads into a fixed ShortArray; never written to disk
val ringBuffer = CircularShortBuffer(RING_BUFFER_SAMPLES)
```

Switch to 16kHz only after detection, for the STT capture window.

### Inference loop

```kotlin
while (isListening) {
    audioRecord.read(chunk, 0, CHUNK_SIZE)
    ringBuffer.write(chunk)
    val features = preprocessToMelSpectrogram(ringBuffer.snapshot())
    val confidence = ortSession.run(inputTensor(features)).confidence
    if (confidence > DETECTION_THRESHOLD) {
        onDetected()
        break
    }
}
```

Target: ~10 inferences/second at 8kHz. On NNAPI, this should consume <5% of a single CPU core. Measure with Android Battery Historian before shipping.

### Detection handoff

```kotlin
private fun onDetected() {
    cuePlayer.playStartListeningCue()  // ToneStartListeningCuePlayer — already exists in core:voice
    voiceInputController.startListening(VoiceCaptureMode.AlertCommand)  // AlertCommand already in the enum
    scope.launch {
        voiceInputController.events
            .filterIsInstance<VoiceInputEvent.Transcript>()
            .first()
            .let { event -> voiceCommandService.handle(event.text) }
    }
}
```

`VoiceCaptureMode.AlertCommand` already exists in the enum for exactly this case. `VoiceCommandService` handles the complete QIR routing → skill execution → TTS confirmation pipeline with no changes needed.

### VAD endpointing

Use Sherpa-ONNX VAD from #821 to detect end-of-utterance after wake word, rather than a fixed silence timeout. This avoids cutting off mid-sentence. Initially can ship with a 1.5s silence timeout as a fallback while #821 is pending.

### Settings toggle

Add "Listen for Hey Jandal" toggle in Settings → Voice (existing section). Backed by `ChatPreferences` DataStore (or a new `VoicePreferences` key). When toggled on and `VoiceInteractionService` is not the default assistant, prompt the user to complete the role assignment first (#983 Settings UX).

### Microphone contention

```kotlin
audioManager.requestAudioFocus(
    focusRequest = AudioFocusRequest.Builder(AUDIOFOCUS_GAIN_TRANSIENT).build()
)
// On AUDIOFOCUS_LOSS: pause ring buffer capture
// Resume on AUDIOFOCUS_GAIN
```

On `VoiceInteractionService.onShutdown()` or toggle-off: release `AudioRecord`, cancel coroutines, stop inference loop cleanly.

## Acceptance criteria

- Saying "Hey Jandal" from any screen (including when the app is backgrounded) triggers the haptic + audio chime and STT capture
- Resulting transcript routes through `VoiceCommandService` → QIR → skill → TTS confirmation identically to widget/Quick Actions voice
- Service pauses automatically during phone calls, camera use, and other exclusive audio focus holders
- Settings toggle persists across reboots
- Battery impact: <5% additional drain over 8 hours standby — verify with Battery Historian on S23 Ultra
- Tested on S23 Ultra; NNAPI delegation confirmed active for the inference loop

## Related

- Parent: #65 — Hey Jandal / Android Assistant epic
- Depends on: #983, #984, #821
- Follow-on: false positive tuning (secondary verification pass)
- Shared infrastructure with: #64 — Live Mode (ONNX Runtime + VAD)

lokhor:
## Summary

Combines #983 (Default Assistant) and #985 (Hey Jandal wake word) into one PR since they share the same manifest/voice infrastructure foundation.

Closes #983
Closes #985
Closes #988

---

## fix(#983): Jandal as Android Default Digital Assistant

Fixes the Samsung One UI bug where pressing "Set" produced no visible result — Jandal now sticks as the selected digital assistant and `voice_interaction_service` is properly written.

### Root causes found and fixed (in order)

| Commit | Cause | Fix |
|---|---|---|
| `6f5510a5` | Samsung One UI RoleManager picker returns `RESULT_CANCELED` | Fall back to `Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS` with Snackbar |
| `3e6b6115` | `android:permission` on `AssistActivity` blocked OS eligibility validation | Removed manifest permission gate; added runtime `intent.action` check |
| `ef2c5398` | Trampoline `AssistActivity` calls `finish()` in `onCreate()` — OS treats it as no-op | Deleted `AssistActivity`; added `ACTION_ASSIST` intent-filter directly to `VoiceCommandActivity` |
| `be172b39` | `AssistantRoleBehavior.isPackageQualified()` rejects packages without a `RecognitionService` | Added `JandalRecognitionService` stub + `android:recognitionService` in `interaction_service_config.xml` |

The final log evidence: `W AssistantRoleBehavior: Package com.kernel.ai.debug has an unqualified voice interaction service` — the OS was accepting the role grant, then immediately reverting it.

### JandalRecognitionService

A stub `RecognitionService` — never invoked for actual speech recognition (Jandal's STT runs via Sherpa-ONNX). Returns `ERROR_CLIENT` if called. Exists solely to pass the OS package qualification check at role-grant time.

---

## feat(#985): Hey Jandal wake word detection service infrastructure

Always-on "Hey Jandal" detection. Gracefully disabled when model file is absent — dropping `hey_jandal.onnx` into `app/src/main/assets/models/wakeword/` activates it with no code change.

### What ships

**`core:voice`**
- `WakeWordDetector` — interface: `start(onDetected)`, `stop()`, `isAvailable`
- `WakeWordPreferences` — DataStore: `heyJandalEnabled`, `confidenceThreshold`
- `OnnxWakeWordDetector` — 3-stage openWakeWord pipeline via ONNX Runtime; ring buffer AudioRecord at 16kHz; no-op when model absent

**`app`**
- `WakeWordService` — `foregroundServiceType="microphone"` foreground service; on detection: STT → `VoiceCommandService.handle()` → re-arm
- `KernelAIApplication` — observes `heyJandalEnabled` and starts/stops `WakeWordService`

**`feature:settings`**
- "Hey Jandal" section in VoiceScreen: assistant role badge + "Listen for Hey Jandal" toggle
- Toggle disabled with contextual message when model absent or Jandal not set as default assistant

## Detection pipeline

```
AudioRecord (16kHz mono)
  → ring buffer
  → OnnxWakeWordDetector (3-stage openWakeWord)
  → confidence ≥ threshold
  → VoiceInputController.startListening()
  → VoiceCommandService.handle()
  → QIR routing → skill execution → TTS
  → re-arm
```

### Default assistant (#983)
- [ ] Settings → Default Apps → Digital assistant → select Jandal → stays selected (no revert)
- [ ] `settings get secure voice_interaction_service` returns `com.kernel.ai.assistant/com.kernel.ai.assistant.JandalVoiceInteractionService`
- [ ] Side key shortcut opens Jandal voice overlay
- [ ] VoiceScreen shows "Jandal is your default assistant" badge

### Wake word (#985 — requires model file)
- [ ] Settings → Voice → Hey Jandal section visible
- [ ] Toggle disabled with "model not yet available" when `hey_jandal.onnx` absent
- [ ] After dropping model: toggle enables, notification appears when on
- [ ] "Hey Jandal" triggers cue + STT capture + command execution
- [ ] Service pauses on mic focus loss (call, camera)
- [ ] Toggle off stops service and removes notification


NickMonrad:
Set button doesn't do anything in voice settings 

Logs: kernel_debug_log_0.1.0_20260528_083252_875.txt

lokhor:
**Root cause found and fixed** — pushed as `45d4b6cd`.

**What was happening:**

`RoleManager.createRequestRoleIntent(ROLE_ASSISTANT)` returns `null` when the system hasn't yet indexed the app as a valid voice interaction service. This is common on first install (the package manager scan hasn't completed) and on Samsung One UI (stricter `RoleManager` implementation). The previous code silently no-oped on null — button tap, nothing happens, no log output.

The log confirms this: the button tap registered (`ViewPostIme pointer 0/1`) but there are zero `KernelAI` log lines, meaning the null branch was taken silently.

**Fix:**

Try `createRequestRoleIntent` first. If it returns null, fall back to `Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS` — this always works and takes the user directly to the right screen where they can select Jandal as the default assistant manually.

A `KernelAI` log line now distinguishes which path was taken, so future debugging is straightforward.

**To test:** Install the new APK (CI will build from the updated branch), tap Set — either the system role dialog appears directly, or Default Apps settings opens. Both outcomes allow setting Jandal as the assistant.

lokhor:
**Root cause found and fixed** (commit `6f5510a5`)

The Set button *was* launching `RoleManager.createRequestRoleIntent(ROLE_ASSISTANT)` successfully (log confirmed: `VoiceScreen: launching RoleManager assistant request`). On Samsung One UI 8.0, the role picker Activity starts but **immediately finishes with `RESULT_CANCELED` showing zero UI** — Jandal doesn't appear in Samsung's internal assistant eligibility check, so the dialog never renders.

The previous `ActivityResultCallback` was a no-op comment ("Result is ignored"), so the user saw nothing and the button appeared broken.

**Fix:** The callback now calls `isRoleHeld()` after the picker returns. If the role still isn't held (i.e. picker silently dismissed — the Samsung case), it automatically opens `Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS` where the user can tap **Digital assistant** and select Jandal manually. This gives a reliable path on all OEMs.

Also updated the supporting text to say "Opens Default Apps → Digital assistant" so users know exactly what to look for.

app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
```diff
@@ -0,0 +1,231 @@
...
+                    }
+                    else -> Unit
+                }
+            }
+        }
+
+        return START_STICKY
+    }
+
+    override fun onDestroy() {
```
lokhor, line 106: **[P1] Leaked transcript waiter on non-transcript outcomes**

`handleDetection()` filters only for `VoiceInputEvent.Transcript`, but alert-command sessions can legally end with `Error` or `ListeningStopped` (timeout, recogniser failure, permission denial). When that happens this coroutine never completes and stays subscribed to the shared `voiceInputController.events` flow indefinitely. The next real transcript from any later voice session will wake every leaked waiter and route the same text through the handler multiple times.

Add a `takeWhile` or `first { it is Transcript || it is Error || it is ListeningStopped }` guard so the collector always terminates on session end.

feature/widget/src/main/java/com/kernel/ai/feature/widget/VoiceCommandActivity.kt
```diff
@@ -61,10 +66,28 @@ class VoiceCommandActivity : ComponentActivity() {
...
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
+        if (ContextCompat.checkSelfPermission(this, Manifest.permission.RECORD_AUDIO)
+                != PackageManager.PERMISSION_GRANTED) {
+            // Permission missing — request it. The system dialog will appear over this
+            // translucent activity. On grant, startVoiceSession(); on deny, finish().
+            requestMicPermission.launch(Manifest.permission.RECORD_AUDIO)
+            return
```
lokhor, line 84: **[P2] Wake word stays paused when RECORD_AUDIO is denied**

`JandalVoiceInteractionSession.onShow()` pauses `WakeWordService` before launching `VoiceCommandActivity`. The new permission-denial path calls `finish()` immediately — no voice session starts, so `WakeWordService` never sees a `ListeningStopped` event and never resumes. "Hey Jandal" detection stays dead until the service restarts.

Call `WakeWordService.resume()` before or alongside `finish()` on the denial branch.

feature/widget/src/main/java/com/kernel/ai/feature/widget/WidgetNavigator.kt
lokhor, line 1: **[P2] Repeated identical chat handoffs silently dropped**

`navigateToChat` now reuses the existing `MainActivity` via `SINGLE_TOP|CLEAR_TOP`, but `adbChatInput` is still a plain `mutableStateOf<String?>`. If the same transcript arrives twice, Compose suppresses the second state assignment and `LaunchedEffect(initialChatQuery)` never re-runs — the repeated handoff is silently dropped. Apply the same `QuickActionRequest`-style serial treatment used for `quick_action_input`, or leave `navigateToChat` using `FLAG_ACTIVITY_NEW_TASK` alone so each call creates a fresh delivery.

feature/settings/src/main/java/com/kernel/ai/feature/settings/VoiceScreen.kt
lokhor, line 1: **[P2] Hey Jandal toggle doesn't check RECORD_AUDIO before enabling**

The toggle persists `heyJandalEnabled=true` immediately and `KernelAIApplication` responds by starting `WakeWordService` — but `RECORD_AUDIO` is never checked or requested here. On a fresh install a user can enable "Listen for Hey Jandal" without mic permission; `OnnxWakeWordDetector` then tries to open `AudioRecord`, fails silently, and the service idles forever with detection broken. Check `RECORD_AUDIO` is granted before writing the preference, or disable the toggle UI until permission is already held.

lokhor (commented):
4 findings — 2 need fixing before merge, 2 are lower priority but worth addressing:

**Must fix:**
- **[P1] Leaked transcript waiter** in `WakeWordService.handleDetection()` — stays subscribed forever on timeout/error, causing phantom double-routing on the next real query.
- **[P2] Wake word stays paused after RECORD_AUDIO denial** — `VoiceCommandActivity` calls `finish()` but never calls `WakeWordService.resume()`, so Hey Jandal detection is dead until the service restarts.

**Should fix:**
- **[P2] Repeated identical chat handoffs dropped** — `adbChatInput` has the same repeated-key problem that was just fixed for `quick_action_input`; `navigateToChat` with `SINGLE_TOP|CLEAR_TOP` will silently drop a second identical query.
- **[P2] Hey Jandal toggle missing RECORD_AUDIO gate** — enabling the toggle without mic permission leaves `WakeWordService` running with a broken detector.

Fixing all four now.

lokhor:
All 4 review findings addressed in `64d8131b`:

- **[P1] Leaked transcript waiter** — `handleDetection()` now uses `.first { Transcript || Error || ListeningStopped }` so the collector always terminates; re-arms on non-transcript outcomes.
- **[P2] Wake word stays paused on permission denial** — `JandalVoiceInteractionSession.onHide()` now unconditionally calls `WakeWordService.resume()`. Covers all early-exit paths from `VoiceCommandActivity`.
- **[P2] Repeated chat handoffs dropped** — reverted `navigateToChat` to `FLAG_ACTIVITY_NEW_TASK` only; each call delivers a fresh intent.
- **[P2] Hey Jandal toggle missing RECORD_AUDIO gate** — `VoiceScreen` now checks permission before writing the pref; requests `RECORD_AUDIO` if missing; enables on grant, leaves toggle off on denial.

Requesting scoped re-review of `64d8131b` only.

lokhor:
Fixed in commit `461ee5db`.

**Root cause:** Samsung One UI 8 overrides `RoleManager.ROLE_ASSISTANT` with a proprietary `RoleControllerService` that silently rejects third-party `VoiceInteractionService` packages — the standard role picker dialog never appears, so "Set" appeared to do nothing.

**Fix:** On Samsung (detected via `Build.MANUFACTURER`), the Set button now launches `Settings.ACTION_VOICE_INPUT_SETTINGS` directly, which opens Samsung's native "Default digital assistant" chooser page — the one that Samsung actually honours for third-party apps. Falls back to `ACTION_MANAGE_DEFAULT_APPS_SETTINGS` if that specific page isn't available. All other OEMs keep the existing `createRequestRoleIntent(ROLE_ASSISTANT)` flow.

**To test:** Settings → Voice → tap "Set". On S23 Ultra you should land on the Samsung Default digital assistant chooser immediately. Select Jandal, return to the app — the badge should update to "Jandal is your default assistant" on the next screen resume.

Issue #986, opened 2026-05-25
feat: wake word false positive reduction — secondary STT verification


lokhor:
## Summary

Reduce false positive activations after the wake word service (#985) ships and real-device testing reveals the false positive rate in normal NZ speech use.

**Gate:** do not implement until #985 is shipped and real-device testing on S23 Ultra shows a false positive rate above ~2 activations/hour in normal daily use. The model training pipeline (#984) should have already tuned `false_positive_weight` to make this unnecessary, but this issue exists as the follow-on if it isn't enough.

## Design

When `WakeWordDetector` confidence crosses a **lower secondary threshold** (below the primary activation threshold), rather than immediately triggering the full pipeline:

1. Capture the audio window (the 3-second ring buffer contents at detection time)
2. Run it through a lightweight STT pass using the existing `VoiceInputController` (Vosk or native Android — already in `core:voice`)
3. Check if the transcript contains "hey jandal" (case-insensitive, allowing for phonetic near-misses)
4. Only if the verification passes: trigger `onDetected()` and hand off to `VoiceCommandService`

This adds ~150–200ms latency on true positives but reduces false positives by ~80% per community validation (see #65 comment from @m13v — same approach validated in production voice-first agents).

## Dual-threshold approach

```
confidence < LOW_THRESHOLD  → ignore
LOW_THRESHOLD ≤ confidence < HIGH_THRESHOLD → secondary STT verification
confidence ≥ HIGH_THRESHOLD → activate immediately (no verification delay)
```

This keeps the fast path fast for high-confidence detections while catching borderline cases.

## Related

- Parent: #65 — Hey Jandal / Android Assistant epic
- Depends on: #985 (wake word service must ship first; real-device false positive data needed before implementing)

NickMonrad:
Backlog grooming: reclassified `launch:blocking` → `launch:post`. The concrete wiring is tracked in #996 (which carries `launch:blocking`). This issue remains the **post-implementation tuning/measurement** follow-up — adjust thresholds only after #996 ships and real-device FP-rate data is collected.

NickMonrad:
Closing as completed/superseded by #996.

#986 defined the dual-threshold secondary STT verification pattern for low-confidence wake-word detections. #996 implemented the concrete wiring by passing the Sherpa-ONNX `transcribeBlocking` / `containsWakePhrase` verifier into `WakeWordService` as the `verifyWindow` callback, and #996 is now closed as completed.

[...]

PR #995, opened 2026-05-27
feat(#821): Sherpa-ONNX streaming STT engine (Zipformer int8)

### LINKED ISSUE #821: research: Sherpa-ONNX / Sherpa-ncnn STT + VAD evaluation
## Summary

We're already shipping Sherpa-ONNX for TTS (Sherpa-ONNX Piper voices). This issue tracks evaluating Sherpa-ONNX and Sherpa-ncnn as a native STT and VAD stack — keeping the whole voice pipeline inside one ecosystem we already depend on.

## Motivation

Current STT stack:
- **Vosk** — offline, works, but weak on NZ accent and dated models
- **Android native STT** — excellent recognition (#678, done in PR #714), but requires network on some devices and no offline guarantee

Sherpa offers offline streaming STT with modern CTC/RNN-T models on Android, plus VAD (voice activity detection) that would enable cleaner endpointing and a path to wake word detection — without adding a new SDK dependency.

### Track 1 — Sherpa-ONNX streaming STT

- Repo: https://github.com/k2-fsa/sherpa-onnx
- Kotlin API available (same ecosystem as TTS)
- Models to evaluate:
  - **Paraformer** (streaming, zh+en) — fast, good accuracy
  - **Zipformer** (streaming, en) — strong English, comparable to Whisper
  - **Moonshine** (streaming, en) — small/fast, designed for edge
  - **Kaldi-native-fbank** models — established baseline
- Check SM8550 (Snapdragon 8 Gen 2) NNAPI/QNN delegate compatibility
- Compare WER on NZ English phrases vs Vosk baseline

### Track 2 — Sherpa-ncnn (alternative runtime)

- Repo: https://github.com/k2-fsa/sherpa-ncnn
- Uses ncnn instead of ONNX Runtime — potentially lighter, no ONNX Runtime dependency
- Supports streaming STT + VAD + TTS on Android arm64
- Kotlin bindings available
- Evaluate as lighter alternative if ONNX Runtime AAR size is a concern

### Track 3 — Sherpa-ONNX VAD

- Docs: https://k2-fsa.github.io/sherpa/onnx/vad/index.html
- Supports **silero-vad** (MIT) and **ten-vad** (Apache 2.0 variant)
- Android examples available
- Use cases:
  - Better push-to-talk endpointing (stop listening when speech ends)
  - Voice slot-fill retry on silence (#790)
  - Foundation for wake word (#65) — VAD gates the keyword detector
  - Start-listening cue (#791) — VAD confirms mic is hot

### Track 4 — Wake word path

- Sherpa-ONNX supports keyword spotting models alongside VAD
- Evaluate as a replacement/alternative to Picovoice Porcupine (#65)
- Keeps wake word fully offline with no per-activation API cost

## Integration notes

- We already ship the Sherpa-ONNX AAR for TTS — STT models can use the same AAR
- STT controller should slot in behind `VoiceInputController` seam (same as Vosk / Android native)
- VAD can be a separate lightweight component owned by `VoiceInputController` implementations
- Model files are gitignored; `download-models.sh` should be extended to cover STT models

## Acceptance criteria (research phase)

- [ ] WER comparison: Sherpa-ONNX STT vs Vosk vs Android native on NZ English test phrases
- [ ] Latency: TTFT (time to first transcript token) on S23 Ultra for streaming models
- [ ] APK size impact: Sherpa-ONNX STT AAR overhead (already included for TTS — may be zero)
- [ ] VAD accuracy: silero-vad endpointing tested in push-to-talk flow
- [ ] Wake word feasibility assessment documented

## Related

- #678 — Android native STT (done, PR #714)
- #700 — Parakeet CTC evaluation (may be superseded by this)
- #703 — Whisper.cpp vs Vosk evaluation (may be superseded by this)
- #790 — Voice slot-fill retry on no-speech (VAD helps here)
- #791 — Start-listening cue (VAD helps here)
- #65 — Wake word (VAD + keyword spotting path)
- #350 — Phase 3F: Voice Interface (parent)

### LINKED ISSUE #1005: Assistant, wakeword and widget microphone don't work on Huawei Honor Magic 8 Pro
https://www.gsmarena.com/honor_magic8_pro_5g-14231.php

Tested with Android STT enabled and the following do not work:
-Widget microphone button
-setting default assistant app (button does nothing)
-long press power button to trigger assistant
-wakeword 'hey jandal'
-verbose logging doesnt export any logs 

it might be a device permissions issue

Vosk works for normal mic button TTS

lokhor:
## Summary

Implements Sherpa-ONNX as a fully offline, on-device STT engine behind the existing `VoiceInputController` seam. Sherpa-ONNX is already a runtime dependency (used for Piper TTS), so there is zero APK size increase from adding STT.

Also absorbs the fixes from PR #1006 (Honor Magic 8 Pro — STT backend, assistant role button, log export), which was merged directly into this branch at commit `32e51042` rather than landing via `main` first.

Closes #821

### `SherpaOnnxVoiceInputController` (new)
- Reflection-only — no `com.k2fsa.sherpa.onnx.*` compile-time imports; `:core:voice` compiles without the AAR
- Streaming Zipformer int8 model (`sherpa-onnx-streaming-zipformer-en-2023-02-21`)
- `AudioRecord` 16 kHz mono PCM, 100 ms chunks on `Dispatchers.IO`
- Drain loop: `while (isReady) decode` — avoids frame drops from single-`if` pattern
- `inputFinished()` + drain before every `getResult()` — flushes Sherpa's tail frames so last words aren't truncated
- Correct `createStream("")` signature (Sherpa 1.13.0 takes a hotwords `String`)
- `reset()` invoked as `recognizer.reset(stream)` — correct receiver
- Empty endpoint (silence-triggered `rule1`) resets stream and continues; timeout flushes before emitting
- Session lifecycle serialised with `Mutex` — no race between concurrent `startListening`/`stopListening`
- Recogniser init serialised with a separate `Mutex` — prevents concurrent native object leaks
- `isAvailable()` guard: returns `Unavailable` when model files absent

### `VoiceInputEngine`
- New `SherpaOnnx` entry (display name: "Sherpa-ONNX (Local)")
- Vosk no longer marked "Recommended" — Sherpa is the better option when the model is present

### `SelectableVoiceInputController`
- Added `SherpaOnnxVoiceInputController` constructor param and `when` branch
- `stopInactiveControllers` / `stopAllControllers` updated

### `KernelModel` / `VoiceViewModel` / `VoiceScreen`
- On-device STT model download (4 HuggingFace files, ~72 MB total)
- Download / cancel / progress wired in `VoiceViewModel`
- `SherpaOnnxSttDownloadCard` composable with Download / Cancel / Delete actions
- `deleteSherpaOnnxStt()` reverts selected engine to Vosk on delete

### Wake word + voice retry (#790)
- `WakeWordService`: 1 STT retry before re-arming; `isHandlingDetection` gate prevents observer double-rearm; `rearmDetector()` in `finally` — always called exactly once
- `ChatViewModel`: 1 retry on Error for `VoiceCaptureMode.Command`
- `VoiceCommandActivity`: 1 retry on blank Transcript and on Error before `finish()`; early `finish()` if retry `startListening()` returns non-`Started`

### Various fixes
- All-caps STT output: `.lowercase(Locale.ROOT)` in `resultText()`
- Wake bloop not playing: `cuePlayer.playCue()` called after `startListening()` success, not via SharedFlow timing
- Wake phrase regex: `\b` word boundaries to prevent false positives on substrings
- Detection sensitivity slider: supporting text explaining higher % = stricter threshold
- `VoiceInputStartResult` import missing in `VoiceCommandActivity`

### From PR #1006 (Honor Magic 8 Pro fixes — Closes #1005)
- `NativeAndroidVoiceInputController`: always start on `OnDevice` backend; Platform fallback retained as safety net
- `VoiceScreen`: Honor-specific deep-link to `ACTION_MANAGE_DEFAULT_APPS_SETTINGS` for set-assistant flow
- `AboutViewModel`: blank logcat detection + stderr capture; surfaces `ExportState.Error` with ADB fallback command instead of sharing empty file

### `scripts/download-models.sh` (new)
```bash
./scripts/download-models.sh stt       # downloads Zipformer int8 archive from k2-fsa releases
./scripts/download-models.sh wakeword  # downloads openWakeWord pipeline models
./scripts/download-models.sh           # both
```

## Model setup

```bash
./scripts/download-models.sh stt
# installs: app/src/main/assets/models/stt/{encoder,decoder,joiner}-epoch-99-avg-1.int8.onnx + tokens.txt
```

## Testing

- `SherpaOnnxVoiceInputControllerTest` — 4 tests covering `isAvailable()` false cases and `startListening()` → `Unavailable` when models absent or Sherpa class not on JVM classpath
- `SelectableVoiceInputControllerTest` — updated to include `SherpaOnnxVoiceInputController` mock
- `VoiceViewModelTest` — STT download/cancel/delete state coverage
- `NativeAndroidVoiceInputControllerTest` — Honor backend fix coverage
- `AboutViewModelTest` — blank logcat and error state coverage
- All unit tests pass; `assembleDebug` clean

## Next steps (follow-up)

- Wire `SherpaOnnxVoiceInputController` as the `verifyWindow` callback in `WakeWordService` for false-positive reduction (issue #986)
- Evaluate NNAPI delegate (`provider = "nnapi"`) for NPU acceleration on S23 Ultra once end-to-end testing is complete
- WER comparison: Sherpa vs Vosk on NZ English phrases (manual device testing)
- Make Sherpa-ONNX the default engine, remove Vosk (issue #1008)


lokhor:
## Code Review — oracle agent

**Verdict: Request changes. Do not merge yet.**

The optional-runtime Sherpa integration is architecturally sound, but there are merge-blocking lifecycle/concurrency bugs and the download script doesn't provide the integrity guarantees it claims.

---

### 🔴 Critical (must fix before merge)

**1. Session lifecycle is not actually serialized — overlapping recordings are possible**
(`SherpaOnnxVoiceInputController.kt:132-173, 193-205`)

`startListening()` calls `activeJob?.cancel()` but the audio loop only exits on the `isRecording` volatile flag — it does not observe coroutine cancellation. `startListening()` never clears `isRecording` or `join()`s the old job before launching a new one. `stopListening()` also mutates `isRecording` and `activeJob` **without holding `sessionMutex`**, so a concurrent stop/start can null out or cancel the wrong job. Two recording loops can overlap against a singleton recognizer and microphone.

**Fix:** One serialized state machine. Capture the old job under lock, set the stop flag, `cancelAndJoin` it, then start the replacement. Keep all `activeJob` mutations under the same mutex.

---

**2. `download-models.sh` claims SHA-256 verification but doesn't do it**
(`scripts/download-models.sh:9, 75-116, 151-160`)

The script header says "SHA-256 hashes are verified" but the STT archive path never supplies an expected hash, and the wake-word downloads explicitly pass `"skip"`. The script also trusts a cached `/tmp/$STT_ARCHIVE` without re-verifying it before extraction. This is a supply-chain integrity hole.

**Fix:** Add real expected SHA-256 values for every artifact and verify cached archives before extraction, or remove the verification claim from the documentation.

---

**3. `AudioRecord.read()` error codes spin forever without surfacing failure**
(`SherpaOnnxVoiceInputController.kt:205-206`)

`if (n <= 0) continue` swallows negative error codes (`ERROR_DEAD_OBJECT`, `ERROR_INVALID_OPERATION`, etc.). A dead recorder produces a tight spin loop that never emits `Error`/`ListeningStopped` and burns CPU indefinitely.

**Fix:** Treat `n < 0` as terminal. `ERROR_DEAD_OBJECT` could optionally trigger recorder recreation; other error codes should break the loop and emit an error event.

---

### 🟡 Major (should fix)

**4. `transcribeBlocking()` leaks native `OnlineStream` on exception**
(`SherpaOnnxVoiceInputController.kt:321-337`)

`mStreamRelease!!.invoke(stream)` is only on the happy path. If `acceptWaveform`/`inputFinished`/`decode`/`resultText` throw, the stream leaks. Once wired into the wake-word verification path, repeated failures accumulate native stream leaks.

**Fix:** Wrap stream lifetime in `try/finally`, same as `audioLoop()` already does correctly.

---

**5. `setProperty()` fails open — silent misconfiguration if Sherpa API drifts**
(`SherpaOnnxVoiceInputController.kt:521-544`)

Required recognizer config fields (encoder/decoder/joiner paths, sample rate) are set via `setProperty()`, which logs-and-continues on failure. If the Sherpa AAR version changes its API, the recognizer can be constructed with all-default config silently, producing garbage results instead of failing fast.

**Fix:** For required STT config fields, let the exception propagate out of `initRecognizer()`.

---

**6. Test coverage doesn't cover the highest-risk behavior**
(`SherpaOnnxVoiceInputControllerTest.kt`)

The 4 tests only cover missing-assets / class-not-found cases. Not covered: session replacement races, `stopListening()` under concurrent access, timeout handling, `AudioRecord.read()` error codes, `transcribeBlocking()` stream cleanup. The `SelectableVoiceInputController` stop test also never verifies Sherpa is actually stopped.

---

### 🔵 Minor notes

- **`ListeningStarted` emitted before mic is open** — event fires before `AudioRecord.startRecording()`, violating the existing cue-player contract.
- **Wake-phrase regex has no word boundaries** — `Regex("(?:hey|a)\\s*(?:jandal|...)")` matches `"a handover"` and `"hey jandalism"`. Add `\b` anchors before using this as a false-positive suppression gate.
- **`ensureRecognizerBlocking()` KDoc overstates safety** — `runBlocking` inside is not safe from Main or arbitrary coroutine dispatchers; "any thread" should be scoped to "dedicated plain threads only".
- **`tryEmit()` + small buffer can silently drop events** — same pattern as existing controllers, but Sherpa's 100 ms partial cadence makes backpressure more likely. Terminal events (`Transcript`, `ListeningStopped`, `Error`) should be lossless.

---

### ✅ What's done well

- Reflection handles are cached outside the hot path — no repeated `Class.forName()` / `getDeclaredMethod()` cost per chunk.
- `audioLoop()` releases the Sherpa stream in a `finally` block correctly.
- `SelectableVoiceInputController` wiring is complete across constructor, engine switch, and all stop-helper paths.
- `VoiceInputEngine.SherpaOnnx` won't break DataStore ordinal compatibility — preferences persist `engine.name`, not ordinal.
- `isAvailable()` provides a clean early unavailable path without touching reflection.

PR #1000, opened 2026-05-28
feat(#984): Hey Jandal wake word — trained model + ML training infrastructure

### LINKED ISSUE #984: feat: local ML training pipeline — custom Hey Jandal wake word model
## Summary

Set up the local ML training pipeline and produce `hey_jandal.onnx` — the custom wake word classifier that runs as the third stage of the openWakeWord inference pipeline. This is an ML/data workstream that runs in parallel with the Android code work (#985).

"Hey Jandal" cannot be sourced from any pre-trained model library — it is an invented word with NZ-specific phonetics (Māori-influenced short front /a/, distinct from both American and British pronunciation). The model must be trained from synthetic data generated locally.

## Platform

**Local machine:** AMD Ryzen 7 5700X3D, AMD Radeon RX 9070 XT (RDNA 4 / Navi 48), Linux.

**ROCm note:** Verify ROCm support for RDNA 4 first — `rocminfo` should list the GPU. If ROCm 6.3+ is available and stable, use it for training. If not, CPU training is the correct fallback — openWakeWord classifiers are tiny (<1MB) and the 5700X3D handles this without issue.

## Architecture — 3-stage ONNX pipeline

openWakeWord is **not** a single monolithic model. It is a fixed 3-stage pipeline:

```
Audio (16kHz, 80ms = 1280 samples per chunk)
  → [Stage 1] melspectrogram.onnx      — raw PCM → mel-spectrogram patch
  → [Stage 2] embedding_model.onnx     — mel patch → 96-dim embedding vector (1 per 80ms frame)
  → ring buffer (last 28 frames = ~2.24s of acoustic context)
  → [Stage 3] hey_jandal.onnx          — 28×96 embedding window → confidence [0, 1]
```

**You only train Stage 3.** Stages 1 and 2 are pre-built shared models provided by the openWakeWord project and must be downloaded separately. All three files must be present at runtime.

### Model files needed at runtime

| File | Source | Size |
|------|--------|------|
| `melspectrogram.onnx` | Download from [openWakeWord releases](https://github.com/dscripka/openWakeWord/releases) | ~170 KB |
| `embedding_model.onnx` | Download from [openWakeWord releases](https://github.com/dscripka/openWakeWord/releases) | ~35 MB |
| `hey_jandal.onnx` | **Produced by this pipeline** | <1 MB |

Place all three in `app/src/main/assets/models/wakeword/`.

### Model contracts (Stage 3 input/output)

The classifier (`hey_jandal.onnx`) receives **pre-computed embeddings**, not raw audio:
- **Input:** `float32[1, 28, 96]` — sliding window of 28 embedding frames from Stage 2
- **Output:** `float32[1, 1]` — confidence that the wake phrase was spoken, in [0, 1]

The backbone (Stage 2) processes 80ms frames at 16kHz (1280 samples/frame) and produces one 96-dim embedding per frame. The classifier sees the most recent 28 frames (~2.24s of context).

**Audio is 16kHz, mono, 16-bit PCM — mandatory and non-negotiable.** This is a hard requirement of the openWakeWord pipeline, not a preference.

### Option A — openWakeWord.com (recommended for first model)

[openWakeWord.com](https://openWakeWord.com) provides a web interface that handles synthetic voice generation (via Piper) and cloud-GPU training, returning a ready-to-use `.onnx` classifier. Fastest path to a working model. Limitations: less control over negative examples and false positive tuning.

### Option B — Google Colab notebook

The [official Colab notebook](https://github.com/dscripka/openWakeWord) walks through synthetic data generation and training step by step. Good for parameter iteration without local GPU setup.

### Option C — fully local (preferred for iteration + privacy)

```bash
# Python 3.10+
pip install openwakeword torch torchaudio onnx onnxruntime
git clone https://github.com/dscripka/openWakeWord
cd openWakeWord

# ROCm path (only if rocminfo confirms RDNA4 GPU):
pip install torch --index-url https://download.pytorch.org/whl/rocm6.3
```

Training is done via the project's Jupyter notebook (`notebooks/training_models.ipynb`), not a CLI script. Run it locally with Jupyter:

```bash
pip install jupyter
jupyter notebook notebooks/training_models.ipynb
```

Key notebook parameters:
- `target_phrase` — the wake phrase string
- `n_samples` — number of synthetic positive examples to generate (500–1000)
- `false_positive_weight` — increase to suppress false positives at the cost of sensitivity (start at 5.0, tune from real-device FP rate in #986)

## Step 1 — Positive example generation: "Hey Jandal"

Generate 500–1000 diverse audio samples. Voice diversity matters more than volume — a model trained on 3 voices will overfit.

**Phonetics to preserve:** "Jandal" = /ˈdʒændl/ — the short front /a/ as in "jam", not the lengthened /ɑː/ common in British or American speech. This is the critical differentiator for NZ recognition.

### TTS sources for NZ-sounding samples

- **Voicebox** ([jamiepine/voicebox](https://github.com/jamiepine/voicebox)) + **Qwen3TTS** — voice cloning from real kiwi speaker recordings; best path to authentic NZ phonetics
- **Sherpa-ONNX VCTK/Semaine voices** — already in project assets; NZ phonetic tuning from #784 feeds directly into sample quality
- **Piper NZ English model** — if available; lightweight and fast
- **Additional diversity:** Coqui XTTS, Kokoro — vary speaker, pitch, and rate

### Augmentation per sample

Convolve with room impulse responses (0.5m–5m simulated distance), mix background noise at −10dB to +5dB SNR, vary playback speed ±20%.

## Step 2 — Negative examples

openWakeWord uses ambient audio as negatives. Essential inclusions:
- `audioset` or `fma` clips — speech and non-speech
- **Similar-sounding phrases** (critical): "hey Daniel", "hey handle", "hey Sandra", "hey gentle", "hey Janet", "hey candle"
- NZ-accent general conversation — the model must not fire on normal NZ speech

## Step 3 — Training (notebook)

Run `notebooks/training_models.ipynb`. The notebook:
1. Calls `openwakeword.utils.get_pretrained_model_paths()` to locate the shared melspectrogram/embedding models (downloads if absent)
2. Computes embeddings for all positive and negative clips using the shared backbone
3. Trains a small FCN classifier (PyTorch) on the pre-computed embeddings
4. Exports the classifier to ONNX via `torch.onnx.export`

The exported file is the Stage 3 classifier only. The shared Stage 1/2 models are managed separately.

## Step 4 — Export and quantisation

The notebook handles the initial ONNX export. After export, optionally quantise:

```python
from onnxruntime.quantization import quantize_dynamic, QuantType
quantize_dynamic("hey_jandal.onnx", "hey_jandal_int8.onnx", weight_type=QuantType.QInt8)
```

Target: <1MB classifier, <5ms total pipeline inference per 80ms chunk on Adreno 740 via NNAPI.

> **Note:** The quantised `hey_jandal_int8.onnx` and unquantised `hey_jandal.onnx` are both valid. The Android detector (`OnnxWakeWordDetector`) loads whichever file is placed at `assets/models/wakeword/hey_jandal.onnx` — rename accordingly.

## Step 5 — Validation before bundling

| Metric | Test | Target |
|--------|------|--------|
| False positive rate | 10 minutes continuous NZ speech | <1 activation/hour |
| True positive rate | 50 manual recordings, primary user | >95% |
| Threshold | Record confidence score distribution on true positives | Document recommended runtime threshold |

FP rate iteration is tracked in #986 (false positive tuning). That issue is gated on real-device measurements from this model.

## Step 6 — Bundle

Place **all three files** in `app/src/main/assets/models/wakeword/`:
- `melspectrogram.onnx`
- `embedding_model.onnx`
- `hey_jandal.onnx`

Document the recommended confidence threshold. The runtime default is configurable in Settings (#985).

## Deliverable

- `hey_jandal.onnx` — validated classifier, bundled with shared backbone models in `app/src/main/assets/models/wakeword/`
- Recommended confidence threshold value documented in a PR comment or `WakeWordPreferences` default
- End-to-end smoke test: device wakes on "Hey Jandal" spoken naturally

## Related

- Parent: #65 — Hey Jandal / Android Assistant epic
- Blocks: end-to-end testing of #985 (code infrastructure exists; model needed for device testing)
- Feeds: #986 — false positive rate tuning (needs real-device FP measurements from this model)
- Related: #784 — Kiwi language corpus tuning (NZ TTS phonetic work feeds positive example quality)

### LINKED ISSUE #985: feat: always-listening wake word detection service

lokhor:
## Summary

Closes #984

Delivers the trained Hey Jandal wake word model and the full ML training pipeline committed to the repo. The `WakeWordService` was previously exiting immediately on startup because the three required ONNX files were absent from assets. They are now bundled.

### Trained model (deployed to assets)

| File | Size | Notes |
|------|------|-------|
| `app/src/main/assets/models/wakeword/hey_jandal.onnx` | 220KB | INT8 quantised Stage 3 classifier |
| `app/src/main/assets/models/wakeword/embedding_model.onnx` | 1.3MB | openWakeWord Stage 2 backbone |
| `app/src/main/assets/models/wakeword/melspectrogram.onnx` | 1.1MB | openWakeWord Stage 1 backbone |

Trained on 954 positive clips: 29 real recordings × augmented + 780 TTS clips across 11 Piper voices and 13 phonetic variants covering the NZ /a/ vowel ("Jandal", "Jandel", "Handal", "Handel" etc).

### ML training infrastructure (committed to `training/`)

| File | Purpose |
|------|---------|
| `training/train.py` | Non-interactive training runner — downloads Stage 1/2 backbones, computes embeddings, trains FCN classifier, exports ONNX, optional `--deploy` to assets |
| `training/wakeword/generate_tts_clips.py` | Piper TTS positive clip generator (multi-voice, augmented) |
| `training/wakeword/generate_hey_jandal.py` | Qwen3-TTS voice-clone clip generator |
| `training/wakeword/README.md` | End-to-end steps: setup → data gen → train → deploy → validate |
| `training/requirements.txt` | Python deps |
| `training/.gitignore` | Excludes `venv/`, `data/`, `piper_voices/`, `output/` |

### `.gitignore` update

Added targeted negation exceptions so the small wake word ONNX files (2.6MB total) are tracked while the global `models/` ignore continues to exclude large LLM weights.

## Testing on device

APK installed to S23 Ultra. To verify:
1. Open **Settings → Voice**
2. **Hey Jandal** toggle should now be enabled (not greyed out / "unavailable")
3. Enable it, then say "Hey Jandal" — voice input should trigger within ~2.5s

## Re-training

If the model needs to be retrained (e.g. after recording more real samples):

```bash
source training/venv/bin/activate
python training/train.py \
    --positive_dir training/wakeword/data/positives \
    --output_dir   training/wakeword/output \
    --epochs       200 \
    --false_positive_weight 5.0 \
    --deploy       app/src/main/assets/models/wakeword
./gradlew installDebug
```

## Related

- Blocks resolved: #985 (OnnxWakeWordDetector code was ready; model was missing)
- FP rate tuning: #986 (needs real-device measurements from this model)


lokhor:
Can we also look at https://github.com/Re-MENTIA/openwakeword-android-kt to compare to our implementation

lokhor:
## Wake word pipeline now working — 62% detection on real recordings

### Root causes found and fixed

**1. PCM scale mismatch (primary cause of all-0 confidence)**
OWW's `melspectrogram.onnx` expects raw int16-as-float32 values (range ±32768). The Kotlin runtime was dividing by 32768 first (range ±1), which shifted mel output by ~90 units — putting every embedding completely outside the trained distribution. Fixed: `chunk[i].toFloat()` instead of `chunk[i] / 32768f`.

**2. OWW augmentation pipeline was missing**
Previous training runs skipped `--augment_clips` (flag errors, missing `torchcodec`). The augmentation step is what creates the RIR-convolved, noise-mixed training features that define the positive class. Without it, features from real inference don't match the training distribution. Fixed: pinned `torchaudio==2.4.0`, installed `torchcodec`.

**3. Real recordings excluded from OWW training**
The 174 real recordings were in `hey_jandal_model/train/positive/` but OWW reads from `hey_jandal_model/hey_jandal/positive_train/`. Fixed: copied real clips into the correct directory before augmentation.

**4. `AudioFeatures.feature_buffer` contamination (`training/train.py`)**
`reset()` clears `raw_data_buffer` but not `feature_buffer`, which is pre-filled with 41 random noise frames on init. This was causing the custom training pipeline to label silence/noise frames as positive.

### Results (chunk-by-chunk streaming, matching runtime behaviour)

| Input | Score |
|-------|-------|
| Real "Hey Jandal" (n=174) | mean=0.58, max=0.94, **62.1% detected at threshold 0.5** |
| Silence | 0.001 |
| Quarantined bad clips | mean=0.31 (acceptable FPR) |

OWW training metrics: accuracy=64.4%, recall=29.6% (on test set), FP/hr=0.27 (target <0.5 ✓)

### Needs owner action
- Connect S23 Ultra via ADB
- `./gradlew installDebug`
- Enable "Hey Jandal" toggle in Settings → Voice
- Say "Hey Jandal" and confirm confidence appears in logcat above threshold

```
adb logcat -s KernelAI | grep -i "wake\|confidence\|jandal"
```

The default threshold is 0.80 — if real-world detection rate feels low, it can be tuned down in Settings or lowered in code (`WAKE_WORD_DEFAULT_THRESHOLD`).

lokhor (commented):
Scoped re-review of fix commit `a4a2239f` only. GitHub will not let me submit `REQUEST_CHANGES` on the author’s own PR, so I’m leaving this as a review comment; blockers remain.

Resolved:
- Security: the exported `VoiceCommandActivity` no longer trusts `EXTRA_PREFILLED_TRANSCRIPT` by itself; it now requires the in-process `WakeWordHandoff.pendingTranscript` match.
- Sherpa: `doInitialize()` and `doInitializeKokoro()` now cover `UnsatisfiedLinkError`, `ExceptionInInitializerError`, and `NoClassDefFoundError`, and Kokoro now preloads ORT first.

Still blocking:
- `VoiceCommandActivity.routePrefilledTranscript()` starts a fresh delayed `lifecycleScope.launch` for every accepted wake-word handoff, but neither `onNewIntent()` nor `routePrefilledTranscript()` cancels the previous job. If trigger B arrives during trigger A’s 400 ms overlay, coroutine A still calls `navigateToActions(..., "A")` and `finish()` first, so the newer transcript is still dropped. The `singleTask`/`onNewIntent` bug is therefore not fully resolved.

New regression from this fix:
- In `handlePrefilledTranscript()`, the mismatch path does `WakeWordHandoff.pendingTranscript = null`. That rejected intent did not own the live token; clearing it lets a stale or external intent invalidate a legitimate pending wake-word handoff before the matching intent arrives.

Please make the handoff consumable per trigger (at minimum, don’t clear on mismatch, and cancel/replace the previous delayed navigation when a newer prefilled transcript is accepted), then I’ll re-check this commit scope.

lokhor (commented):
Reviewed scope: `f91aa7d5` and `9033435c` only.

- **ORT version/linker conflict (`f91aa7d5`)**: **resolved**. Bumping `onnxruntime-android` to 1.24.3, preferring the full Sherpa AAR, and preloading `onnxruntime` before Sherpa class init is the right fix for the `OrtGetApiBase` linker failure. The stale-`UNAVAILABLE` retry in `SherpaOnnxVoiceOutputController` also looks correct.

- **New blocking regression introduced in `f91aa7d5`**: `MainActivity.onResume()` now unconditionally calls `WakeWordService.start()` whenever wake word is enabled. `startForegroundService()` against an already-running service re-enters `WakeWordService.onStartCommand()`, and that method always launches a fresh `voiceInputController.events.collect { ... }` in `serviceScope`. Result: every foreground resume leaks another collector and another detector re-arm path. This needs a guard so the foreground retry only happens when the service is not already running / the earlier background start actually failed.

- **Finding 2 (job cancellation) in `9033435c`**: **still blocking**. `onNewIntent()` now cancels `prefilledNavJob`, but the canceled coroutine in `routePrefilledTranscript()` catches `CancellationException` via `catch (e: Exception)` and still executes `finally { finish() }`. In the rapid re-trigger case, job A cancels, then finishes the activity underneath job B, so the newer transcript can still be lost. The fix needs to avoid finishing from a superseded/canceled job, or otherwise ensure only the currently active navigation job can call `finish()`.

- **Finding 3 (no-clear-on-mismatch) in `9033435c`**: **resolved**. The mismatch branch now preserves `WakeWordHandoff.pendingTranscript`, so an external/stale mismatched intent no longer invalidates the legitimate matching handoff that may arrive next.

lokhor (commented):
Reviewed commit `55c4bcfd` only.

- `WakeWordService.onStartCommand()` collector leak: resolved. The new `eventCollectorJob` guard prevents duplicate `voiceInputController.events.collect { ... }` launches on repeat `onStartCommand()` calls, and the collector is still bound to the service lifetime via `serviceScope.cancel()` in `onDestroy()`.
- `VoiceCommandActivity` canceled-job `finish()` bug: the original finding is resolved for the intended rapid prefilled→prefilled retrigger case. `CancellationException` now escapes instead of falling through to `finish()`.

New blocker introduced in this commit:

`onNewIntent()` still cancels `prefilledNavJob` *before* it knows whether the replacement intent actually contains a prefilled transcript. If a wake-word prefilled overlay is interrupted by a non-prefilled retrigger (for example the assistant entry point or widget mic path), `handlePrefilledTranscript(intent)` returns `false`, no voice session is started on that path, and the canceled job now rethrows without calling `finish()`. That leaves the existing overlay stranded with no navigation and no active STT session. Before this commit, the unconditional `finally { finish() }` at least tore the stale activity down.

So: both original findings are addressed, but this commit introduces a new blocker in the `singleTask` re-entry path and should be followed up before merge.

lokhor (commented):
Reviewed commit `de254067` only.

This resolves the last blocking finding. The cancellation now happens only after `handlePrefilledTranscript()` has confirmed that the incoming extra matches `WakeWordHandoff.pendingTranscript`, so a non-prefilled `singleTask` re-entry no longer tears down the in-flight 400 ms overlay/navigation job. I also traced the supersede path: a newer validated prefilled transcript still cancels the older overlay job before re-routing, which preserves the intended replacement behavior.

I do not see a new blocker introduced by this commit within the reviewed scope.

PR #1105, opened 2026-06-06
chore: remove 88.99 MB .hprof heap dump from git history (#1096)

lokhor:
## Problem

CI's "Publish PR preview APK" stage warns about pushing an 88.99 MB `.hprof` heap dump file during every tag push.

The file was introduced in commit `ec69070` (feat: bulk delete core memories) and was later added to `.gitignore` (line 48) but never cleaned from history.

## Root Cause

CI does `git tag -f "$TAG"` then `git push -f origin "$TAG"`, which pushes all objects reachable from the tag's history — including the stale blob. The blob was also reachable from `ec69070`, which is an ancestor of almost all branches in this repo.

## Fix

Rewrote `ec69070` to remove the `.hprof` file using `git filter-branch --tree-filter --prune-empty`, then cleaned up old refs and garbage collected.

## Verification

- `git ls-tree -r HEAD` no longer contains the blob
- `git rev-list main --objects | grep 6b4d9eab` returns nothing
- The blob is no longer reachable from `main`

## Impact

- Reduces CI tag push payload by ~89 MB
- No functional changes to the codebase


lokhor (commented):
This is a straightforward history rewrite — the PR only removes an 88.99 MB .hprof heap dump blob that was accidentally committed and never cleaned from history. No functional changes to the codebase.

The diff is clean: just `deleted file mode 100644 debug/2026-04-10_21-25-00_299.hprof`. The blob was introduced in ec69070 (feat: bulk delete core memories) and was later added to `.gitignore` (line 48) but never cleaned from history.

CI's tag push (`git tag -f "$TAG"` then `git push -f origin "$TAG"`) was pushing this blob into every release, causing the ~89 MB warning. The fix rewrites ec69070 to remove the blob from its tree, then cleans up old refs and garbage collects.

No security, memory safety, or LiteRT concerns — this is purely a repo hygiene change.

PR #1203, opened 2026-06-12
fix(#747): harden drawer navigation and back handling

### LINKED ISSUE #747: UI bugginess: back button & blank screens — navigation state hardening
## Original report

Unsure how to replicate, but logs attached.

I have found that when using items in the menu - when I press back either in the app or the android back button - sometimes it takes me back to the Chat screen - but nothing is displayed.

I also have found that sometimes it gets stuck in settings where the in-app back button does not do anything, and pressing Android back is closing the application. I had to completely close the app (swipe it away in the app switcher) to make it usable again.

[kernel_debug_log_0.1.0_20260504_065316_800.txt](https://github.com/user-attachments/files/27324625/kernel_debug_log_0.1.0_20260504_065316_800.txt)

---

## Current analysis

This is likely a navigation-state / back-stack consistency bug rather than a rendering bug in Chat itself.

The reported symptoms point at two related failure modes:

1. Menu/drawer navigation followed by back can land the user on the Chats/Chat area with no visible content.
2. Settings can become effectively root-like, where the in-app back button calls `popBackStack()` but nothing happens, and Android back exits the app.

### 1. Drawer/menu destinations are using saved-state restoration too aggressively

Several drawer/menu destinations use the same pattern:

```kotlin
navController.navigate(ROUTE_X) {
    popUpTo(ROUTE_LIST) { saveState = true }
    launchSingleTop = true
    restoreState = true
}
```

This is appropriate for bottom-nav tab restoration, but it is risky for one-off drawer/tool/settings destinations. It can restore stale route state, stale arguments, or nested screen state when the user expects a normal destination.

This is especially suspicious because `navigateToPrimaryRoute()` already contains special handling to avoid saving/restoring transient parameterised routes. Drawer navigation does not currently appear to have equivalent protection.

### 2. Failed `popBackStack()` calls are not handled

Many screens use:

```kotlin
onBack = { navController.popBackStack() }
```

If the current destination has become effectively root-like because of previous `popUpTo` / saved-state restoration interactions, `popBackStack()` can return `false`. In that case the in-app back button appears to do nothing. Android back may then close the activity because there is no meaningful previous destination.

This matches the observed Settings symptom closely.

### 3. Transient routes may be saved/restored from Actions and Chat flows

There are parameterised/transient routes such as:

- `actions?openSheet=...&startVoice=...&widgetQuery=...&widgetVoice=...&draftQuery=...`
- `chat?initialQuery=...&minimalContext=...&speakResponse=...`
- `settings/model_management?scrollTo=...`

Some of these routes include one-shot behaviours guarded by saved-state handles. If a drawer navigation saves/restores them unexpectedly, the app can resurrect a route whose arguments or consumed flags are no longer valid, potentially causing stale, blank, or non-interactive UI.

### 4. Argument-dependent detail routes may render blank if restored without valid args

Some detail routes bail out early if required route arguments are missing or invalid. This is normally fine, but stale saved-state restoration can turn this into a blank screen instead of a graceful fallback.

This is probably secondary to the drawer/back-stack issue, but should be considered during testing.

---

## Recommended implementation approach

Keep this as a focused navigation hardening PR. Do not rewrite the navigation model.

### Scope

1. Add a helper for drawer/menu destinations that does **not** use `restoreState = true` by default.
2. Keep bottom-nav state restoration in `navigateToPrimaryRoute()` unless testing proves it is also faulty.
3. Replace repeated drawer/menu navigation blocks with the safer helper.
4. Add a safe back helper for screens where failed `popBackStack()` should return to the Chats list instead of doing nothing.
5. Add logging or temporary diagnostics if useful to capture route/back-stack transitions during reproduction.
6. Add focused navigation tests where practical.

### Suggested helpers

```kotlin
private fun NavHostController.navigateToDrawerDestination(route: String) {
    val currentBaseRoute = currentBackStackEntry?.destination?.route?.substringBefore('?')
    val targetBaseRoute = route.substringBefore('?')
    if (currentBaseRoute == targetBaseRoute) return

    navigate(route) {
        popUpTo(ROUTE_LIST) {
            saveState = false
        }
        launchSingleTop = true
        restoreState = false
    }
}
```

```kotlin
private fun NavHostController.popBackOrNavigateHome() {
    if (!popBackStack()) {
        navigate(ROUTE_LIST) {
            popUpTo(graph.findStartDestination().id) {
                inclusive = false
                saveState = false
            }
            launchSingleTop = true
            restoreState = false
        }
    }
}
```

The exact implementation may vary, but the principle is:

- bottom nav may preserve tab state;
- drawer/menu destinations should behave like explicit navigation targets;
- failed in-app back should recover to the Chats list rather than silently doing nothing.

---

## Out of scope

- Redesigning the navigation hierarchy.
- Changing bottom-nav routes unless investigation proves they are part of this bug.
- Reworking Settings screen content.
- Changing Chat state, conversation persistence, inference, model loading, routing, tools, memory, permissions, widgets, or prompts.
- Hiding the bug by adding loading placeholders only.

---

## Validation plan

Manual device smoke should cover:

- Chats → drawer/menu → Settings → in-app back → Chats list visible.
- Chats → drawer/menu → Settings → Android back → Chats list or previous expected destination, not app exit unless already at true root.
- Chats → drawer/menu → Notes → back → Chats list visible.
- Chats → drawer/menu → Lists → list detail → back → Lists → back → Chats list.
- Chats → drawer/menu → Clock → back → Chats list.
- Chats → drawer/menu → Convert → back → Chats list.
- Settings → child screen → back → Settings → back → Chats list.
- Actions with sheet/draft/voice params → drawer/menu destination → back.
- Chat opened with initial query → Settings/Model Management → back.
- Rotate screen while on Settings and on a drawer destination, then back.
- Repeat enough times to confirm no blank Chat screen and no stuck Settings state.

Automated validation should include any existing navigation tests and, if cheap, a focused Compose/navigation regression around drawer destination navigation and failed-pop fallback.

---

## Acceptance criteria

- [ ] Pressing in-app back from Settings reliably returns to the expected previous screen or Chats list.
- [ ] Pressing Android back from Settings does not unexpectedly exit the app when Settings was opened from the drawer/menu.
- [ ] Navigating through drawer/menu destinations and then pressing back does not produce a blank Chat/Chats screen.
- [ ] Drawer/menu navigation no longer restores stale transient route state.
- [ ] Bottom-nav behaviour remains intact.
- [ ] Existing Actions, Chat, Tools, Settings, Lists, Notes, Clock, Convert, and model-management navigation flows still work.
- [ ] No inference, model loading, routing, tool execution, memory, permission, widget, prompt, or persistence behaviour changes are introduced.
- [ ] Manual device smoke confirms the original symptoms no longer reproduce.

lokhor:
## Problem

Intermittent navigation/back-stack bugs (#747):
- After using drawer/menu items, pressing back returns to a blank Chat/Chats screen
- Settings can get stuck: in-app back does nothing, Android back exits instead of returning
- Force-close needed to recover

## Root Causes

1. **All 8 drawer/menu destinations** used `saveState = true` + `restoreState = true` — a bottom-nav pattern designed for tab switching. For drawer destinations, this restores stale transient route state and corrupts the back stack when the destination's saved state overlaps with its parent's.

2. **~25 screens** used bare `navController.popBackStack()` with no fallback. When the back stack is exhausted (e.g. Settings as the only entry after drawer navigation), this exits the app or produces blank screens.

### Two new helpers (`KernelNavHost.kt`)

- **`navigateToDrawerDestination(route)`** — no `saveState`/`restoreState`, always pops to `ROUTE_LIST`, idempotent (no-op if already on target).
- **`popBackOrNavigateHome()`** — pops normally; if `popBackStack()` returns false (empty stack), navigates to `ROUTE_LIST` with clean state.

### Drawer navigation (8 items replaced)

Lists, Notes, Clock, Convert, Important Dates, People & Contacts, Meal Plans, Settings — all replaced inline `/popUpTo/launchSingleTop/restoreState` blocks with `navigateToDrawerDestination(...)`.

### Safe-back fallback applied to

Settings and all Settings child screens (UserProfile, ChatPreferences, Memory, Voice, ModelSettings, ModelManagement, About, AppPermissions, ContactAliases), plus all drawer top-level destinations (Lists, Notes, Convert, ImportantDates, SidePanel, MealPlans), detail screens (NoteDetail, ListItems), and ToolsLearn.

### Preserved (unchanged)

- Bottom-nav `navigateToPrimaryRoute()` with transient-param awareness
- `navigateToToolsDestination()`
- Actions route with `openSheet`, `startVoice`, `widgetQuery`, `draftQuery` params and their consumed-state tracking
- Chat `onNewConversation` / `onNavigateToList`
- Widget/ADB query routing
- No inference, model loading, routing, tool execution, memory, permissions, widgets, prompts, or data changes

## Validation

- `./gradlew :app:compileDebugKotlin` ✅ (0 errors, warnings all pre-existing)
- Manual device smoke required per the 11 test scenarios in the issue description

Closes #747

### DIFF
--- a/app/src/main/java/com/kernel/ai/navigation/KernelNavHost.kt
+++ b/app/src/main/java/com/kernel/ai/navigation/KernelNavHost.kt
@@ -175,6 +175,47 @@ private fun NavHostController.navigateToToolsDestination(route: String) {
     }
 }
 
+/**
+ * Navigate to a drawer/menu destination safely.
+ *
+ * Unlike bottom-nav routes, drawer destinations:
+ * - Should NOT use [restoreState] — no stale transient state to restore.
+ * - Should NOT use [saveState] — drawer screens are not tabs.
+ * - Always pop to the Chats list to keep the back stack shallow and predictable.
+ */
+private fun NavHostController.navigateToDrawerDestination(route: String) {
+    val currentBaseRoute = currentBackStackEntry?.destination?.route?.substringBefore('?')
+    val targetBaseRoute = route.substringBefore('?')
+    if (currentBaseRoute == targetBaseRoute) return
+
+    navigate(route) {
+        popUpTo(ROUTE_LIST) {
+            saveState = false
+        }
+        launchSingleTop = true
+        restoreState = false
+    }
+}
+
+/**
+ * Pop the back stack, or navigate to the Chats list if the stack is empty.
+ *
+ * Prevents the app from exiting unexpectedly when the back stack has been
+ * exhausted (e.g. Settings opened from the drawer is the only entry).
+ */
+private fun NavHostController.popBackOrNavigateHome() {
+    if (!popBackStack()) {
+        navigate(ROUTE_LIST) {
+            popUpTo(graph.findStartDestination().id) {
+                inclusive = false
+                saveState = false
+            }
+            launchSingleTop = true
+            restoreState = false
+        }
+    }
+}
+
 @Composable
 internal fun PrimaryBottomBar(
     currentBaseRoute: String?,
@@ -271,11 +312,7 @@ fun KernelNavHost(
                     selected = currentBaseRoute == ROUTE_LISTS,
                     onClick = {
                         coroutineScope.launch { drawerState.close() }
-                        navController.navigate(ROUTE_LISTS) {
-                            popUpTo(ROUTE_LIST) { saveState = true }
-                            launchSingleTop = true
-                            restoreState = true
-                        }
+                        navController.navigateToDrawerDestination(ROUTE_LISTS)
                     },
                     modifier = Modifier.padding(NavigationDrawerItemDefaults.ItemPadding),
                 )
@@ -285,11 +322,7 @@ fun KernelNavHost(
                     selected = currentBaseRoute == ROUTE_NOTES,
                     onClick = {
                         coroutineScope.launch { drawerState.close() }
-                        navController.navigate(ROUTE_NOTES) {
-                            popUpTo(ROUTE_LIST) { saveState = true }
-                            launchSingleTop = true
-                            restoreState = true
-                        }
+                        navController.navigateToDrawerDestination(ROUTE_NOTES)
                     },
                     modifier = Modifier.padding(NavigationDrawerItemDefaults.ItemPadding),
                 )
@@ -299,11 +332,7 @@ fun KernelNavHost(
                     selected = currentBaseRoute == ROUTE_SIDE_PANEL,
                     onClick = {
                         coroutineScope.launch { drawerState.close() }
-                        navController.navigate(ROUTE_SIDE_PANEL) {
-                            popUpTo(ROUTE_LIST) { saveState = true }
-                            launchSingleTop = true
-                            restoreState = true
-                        }
+                        navController.navigateToDrawerDestination(ROUTE_SIDE_PANEL)
                     },
                     modifier = Modifier.padding(NavigationDrawerItemDefaults.ItemPadding),
                 )
@@ -313,11 +342,7 @@ fun KernelNavHost(
                     selected = currentBaseRoute == ROUTE_CONVERT,
                     onClick = {
                         coroutineScope.launch { drawerState.close() }
-                        navController.navigate(ROUTE_CONVERT) {
-                            popUpTo(ROUTE_LIST) { saveState = true }
-                            launchSingleTop = true
-                            restoreState = true
-                        }
+                        navController.navigateToDrawerDestination(ROUTE_CONVERT)
                     },
                     modifier = Modifier.padding(NavigationDrawerItemDefaults.ItemPadding),
                 )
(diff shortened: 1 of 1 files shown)


NickMonrad (approved):
Reviewed PR #1203.

This implementation matches the intended #747 navigation-hardening approach:

- Adds a dedicated `navigateToDrawerDestination(route)` helper that treats drawer/menu routes as explicit destinations and disables `saveState`/`restoreState`, avoiding the bottom-nav restoration pattern that was the main suspect for stale/blank restored state.
- Replaces the repeated drawer item navigation blocks for Lists, Notes, Clock, Convert, Important Dates, People & Contacts, Meal Plans, and Settings with the new helper.
- Adds `popBackOrNavigateHome()` and applies it broadly to Settings, Settings children, drawer destinations, detail screens, Chat, and Tools Learn so failed `popBackStack()` calls recover to the Chats list instead of silently doing nothing.
- Leaves bottom-nav `navigateToPrimaryRoute()`, Actions transient-route handling, widget/ADB query routing, Chat initial-query routing, and model-management arguments intact.
- No inference, model loading, tool execution, memory, permissions, widget, prompt, persistence, or data behaviour changes are introduced.

I did not find a blocking code issue in the diff. The main merge gate should be green CI plus the manual device smoke from #747, especially Settings opened from drawer, Actions with transient params → drawer destination → back, and repeated drawer destination/back cycles.

Approved pending green CI and device smoke.

PR #1204, opened 2026-06-12
fix(#1191): isolate slot-fill state across independent ADB commands

lokhor:
## Root cause

`adbSlotReplyInput` in `MainActivity` was never cleared after consumption. When a new `quick_action_input` intent arrived (via `send_quick_action`) without a `slot_reply_input` extra, the old slot reply value persisted. If the composable re-entered between independent ADB commands (activity going to background and back), `LaunchedEffect(adbSlotReply)` re-fired `onSlotReply()` with the previous case's reply value, overwriting the newly primed pending slot.

This caused a cascade of stale values:

```
Case 1: "set an alarm" + reply "7am" → time=7am ✅
Case 2: "set a timer" + reply "5 minutes" → duration_seconds=7am ❌ (stale from case 1)
Case 3: "open an app" + reply "Spotify" → app_name=5 minutes ❌ (stale from case 2)
Case 4: "send a message" + reply "Mum" → contact=Spotify ❌ (stale from case 3)
Case 6: "add to my list" + reply "eggs" → item=Nick ❌ (stale from case 5)
```

## Fix

**`MainActivity.onNewIntent()`**: Clear `adbSlotReplyInput = null` at the start of `onNewIntent()`, before reading the new intent's extras. This ensures that every independent command starts with a clean slot reply slate. If the new intent carries a `slot_reply_input` extra (legitimate slot reply), it re-sets the value afterward.

**`ActionsViewModel`**: Added `ADB_SLOT_STATE` diagnostic log markers for the full slot-fill lifecycle: `action=create`, `action=reply`, `action=complete`, `action=cancel`.

## S21 Evidence

| Case | BEFORE fix | AFTER fix | Status |
|------|-----------|-----------|--------|
| 1. set an alarm + "7am" | set_alarm (hours/minutes test mismatch) | set_alarm (same) | 🟡 test expectation issue |
| 2. set a timer + "5 minutes" | **duration_seconds=7am** (STALE) | **duration_seconds=5 minutes** ✅ | ✅ Fixed |
| 3. open an app + "Spotify" | **app_name=5 minutes** (STALE) | **app_name=Spotify** ✅ | ✅ **PASSES first time** |
| 4. send a message + "Mum" | **contact=Spotify** (STALE) | contact=Mum ✅ | ✅ Fixed |
| 6. add to my list + "eggs" | **item=Nick** (STALE) | item=eggs ✅ | ✅ Fixed |

`ADB_SLOT_STATE` markers confirm each slot-fill starts with `existingSlots=[]` — no stale state:

```
action=create intent=set_alarm missingSlot=time existingSlots=[]
action=reply intent=set_alarm slot=time reply="7am"
action=complete intent=set_alarm finalSlots=[time]
action=create intent=open_app missingSlot=app_name existingSlots=[]
action=reply intent=open_app slot=app_name reply="Spotify"
action=complete intent=open_app finalSlots=[app_name]
```

## Remaining issues (separate from #1191)

1. **Case 1 test expectation**: Harness expects `hours` and `minutes` in the NativeIntentHandler log marker, but `setAlarm` only outputs `time=7am`. The alarm IS set correctly internally — this is a test-harness assertion bug, not a product bug.
2. **Case 2 param format**: Slot-fill returns raw text `"5 minutes"` as `duration_seconds`. The NativeIntentHandler expects an integer. Duration parsing in the slot-fill path is a separate design issue.
3. **Multi-slot intents (send_sms, send_email, add_to_list)**: The harness sends only one slot reply per test case, but these intents require 2-3 slots. This is a fixture limitation.
4. **#1192**: Parked-location memory QIR gap remains separate.

## Unit tests added

**SlotFillerManagerTest**:
- `sequential slot completions do not leak values between independent requests`
- `current reply value wins over stale existing params`
- `cancel clears all pending state for new independent requests`

**ActionsViewModelTest**:
- `sequential independent slot fills do not leak values`

## Validation

- `:core:skills:testDebugUnitTest` — ✅ 0 failures
- `:feature:chat:testDebugUnitTest` — ✅ 0 failures
- `python3 -m py_compile scripts/*.py scripts/adb_harness/*.py` — ✅ clean
- S21 slot-fill — stale carryover eliminated, ADB_SLOT_STATE markers confirmed
- S21 safe-smoke — #1190 not regressed (same results as pre-#1191)

## Related

- #1190 (command/model state carryover) — fixed by #1202, not regressed
- #1192 (parked-location memory QIR gap) — remains separate

### DIFF
--- a/app/src/main/java/com/kernel/ai/MainActivity.kt
+++ b/app/src/main/java/com/kernel/ai/MainActivity.kt
@@ -125,6 +125,12 @@ class MainActivity : ComponentActivity() {
     override fun onNewIntent(intent: Intent) {
         super.onNewIntent(intent)
         setIntent(intent)
+        // #1191: Clear one-shot ADB extras before reading new intent extras.
+        // When a new `quick_action_input` arrives without `slot_reply_input`, the stale
+        // slot reply value would persist and re-fire the LaunchedEffect in ActionsScreen
+        // if the composable re-enters between independent commands, causing the previous
+        // case's slot reply to overwrite the newly primed pending slot.
+        adbSlotReplyInput.value = null
         intent.getStringExtra("chat_input")?.let { adbChatInput.value = it }
         intent.getStringExtra("quick_action_input")?.takeIf { it.isNotBlank() }?.let {
             val voice = intent.getBooleanExtra("quick_action_is_voice", false)
--- a/core/skills/src/test/java/com/kernel/ai/core/skills/slot/SlotFillerManagerTest.kt
+++ b/core/skills/src/test/java/com/kernel/ai/core/skills/slot/SlotFillerManagerTest.kt
@@ -337,4 +337,97 @@ class SlotFillerManagerTest {
         assertFalse(manager.hasPending)
         assertNull(manager.pendingRequest)
     }
+
+    @Test
+    fun `sequential slot completions do not leak values between independent requests`() {
+        // Simulate: complete open_app(app_name=Spotify), then new set_timer(duration_seconds=5min)
+        // The second completion must not inherit app_name from the first.
+
+        // First request: open_app → fill app_name=Spotify
+        manager.startSlotFill(
+            conversationOne,
+            PendingSlotRequest(
+                intentName = "open_app",
+                existingParams = emptyMap(),
+                missingSlot = SlotSpec("app_name", "Which app?"),
+            ),
+        )
+        val firstResult = manager.onUserReply(conversationOne, "Spotify")
+        assertInstanceOf(SlotFillResult.Completed::class.java, firstResult)
+        val completed = firstResult as SlotFillResult.Completed
+        assertEquals("open_app", completed.intentName)
+        assertEquals("Spotify", completed.params["app_name"])
+        assertFalse(manager.hasPending)
+
+        // Second independent request: set_timer → fill duration_seconds=300
+        manager.startSlotFill(
+            conversationOne,
+            PendingSlotRequest(
+                intentName = "set_timer",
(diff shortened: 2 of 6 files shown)


NickMonrad (approved):
Review approved.

This looks good as a focused fix for the stale ADB slot-reply carryover path. The root cause is plausible and specific: `adbSlotReplyInput` was a one-shot value but was not cleared before subsequent `quick_action_input` intents without a `slot_reply_input` extra. Clearing it at the start of `onNewIntent()` before reading new extras prevents the previous slot reply from re-triggering `LaunchedEffect(adbSlotReply)` during independent ADB commands.

The added `ADB_SLOT_STATE` diagnostics are useful and appropriately scoped to create/reply/complete/cancel. The unit tests cover the key non-leak behaviours in `SlotFillerManager` and `ActionsViewModel`, and CI is green on the current head.

The S21 evidence shows the important #1191 stale value pattern is fixed:

- `set_a_timer` no longer inherits `7am` from the alarm case;
- `open_an_app` now receives `Spotify` rather than stale `5 minutes`;
- `send_a_message` receives `Mum` rather than stale `Spotify`;
- `add_to_my_list` receives `eggs` rather than stale `Nick`.

Non-blocking follow-up: the PR body correctly calls out remaining non-carryover work. Please make sure those do not get lost after merge: alarm marker expectation, timer duration parsing (`5 minutes` → `300`), and multi-slot harness coverage/fixture limitations should either remain under #1191 if that issue stays open, or be split into follow-up issues before #1191 is closed.

Good to merge once you are happy with that tracking.

PR #1208, opened 2026-06-12
fix(#1206): robust chat background import and cleanup

### LINKED ISSUE #1206: Robust chat background management: solid colours, image import, and cleanup
## Summary

Selecting a custom chat background image can crash when the user chooses an image from the Android file picker / Files app rather than the gallery.

The robust fix is to import selected image wallpapers into app-private storage and persist the app-owned URI/path, rather than storing an external provider URI and relying on provider-specific permission behaviour.

This must fit into the existing chat background UX, which includes both **solid colour backgrounds** and **custom image backgrounds**. The fix should treat Default, Solid colour, and Image as distinct background modes rather than assuming the background is always an image.

Because app-owned image imports can create wallpaper files over time, the fix must also include a small user-facing cleanup path for deleting old/unused imported image wallpapers.

## Observed behaviour

- In Chat Preferences / chat background settings, choose a custom background image.
- Selecting from Gallery appears to work.
- Selecting from the file system / Files app crashes the application.

## Existing code context

The current settings screen uses `ActivityResultContracts.GetContent()` and then immediately calls `contentResolver.takePersistableUriPermission(...)` on the returned URI before saving it as `wallpaperImageUri`. `GetContent()` does not guarantee a persistable grant, so Android Files / DocumentsUI can return a URI that throws `SecurityException` when the app attempts to persist permission.

Current persisted wallpaper state is simple DataStore state:

- `wallpaperType`: `none`, `color`, or `image`.
- `wallpaperColor`: ARGB value for solid colour wallpaper.
- `wallpaperImageUri`: external content URI string when image wallpaper is active.

Chat rendering currently fails closed if image wallpaper decode/open fails, but settings selection can crash before that because the picker result path is not defensive.

## Existing UX context

Chat background settings already support more than one type of background:

- Default / no custom background.
- Solid colour background options.
- Custom image wallpaper.

The robust image import fix must not regress the solid colour options. Selecting a solid colour should remain lightweight and should not create image files. Image cleanup controls should apply only to imported image wallpapers, not solid colour choices.

## Likely cause

The current picker path appears to use `ActivityResultContracts.GetContent()` and then attempts to call `contentResolver.takePersistableUriPermission(...)` on the returned URI.

`GetContent()` returns a transient content URI and does not guarantee a persistable URI grant. Some providers, especially Android Files / DocumentsUI depending on the selected source, can return a URI where `takePersistableUriPermission()` throws `SecurityException` because the grant is not persistable.

There is also a longer-term fragility: storing the external provider URI in preferences means the wallpaper depends on that external URI remaining readable after process death, reboot, provider changes, or file moves.

## Goal

Make chat background selection robust across all current background modes:

- Default background.
- Solid colour backgrounds.
- Image wallpaper selected from Gallery.
- Image wallpaper selected from Android Files / file system provider.

For image wallpapers, copy the selected image into app-private storage and store an app-owned reference.

Also provide a clear user interface to remove imported image wallpaper files that are no longer wanted, so the app does not silently accumulate old images in private storage.

## Recommended UX flow

Treat chat backgrounds as a small background picker/manager, not an image-only picker.

Suggested structure in Chat Preferences:

1. A `Chat background` row or section showing the current mode:
   - `Default`
   - `Solid colour`
   - `Custom image`
2. A preview of the current background where practical.
3. A background picker UI, such as a bottom sheet or settings subsection, with these choices:
   - `Default background`
   - `Solid colours`
   - `Choose image`
   - `Previously used images` or `Manage imported images` if app-owned image wallpapers exist.

### Default behaviour

- Selecting `Default background` clears any active solid colour or image selection.
- It should not delete imported image wallpapers unless the user explicitly chooses a delete/remove action.

### Solid colour behaviour

- Solid colour choices should remain first-class background options.
- Selecting a solid colour sets the active background mode to solid colour.
- Solid colours must not be represented as imported image files.
- Solid colours should not be affected by image cleanup actions.
- If a custom image was active before selecting a solid colour, the image can remain in imported image history unless the user deletes it or cleanup removes it as unused.

### Image behaviour

- `Choose image` opens a picker that works with Gallery and Android Files.
- On success, the app copies the selected image into app-private storage, stores the app-owned reference, and sets the active background mode to image.
- On failure, the previous background mode remains unchanged and the app shows a clear non-crashing error.
- If previously imported images are retained, users should be able to reselect them without re-importing.

### Cleanup behaviour

The UI must make the difference clear between:

- `Use default` / `Clear background`: stop using a custom background, but do not necessarily delete stored images.
- `Delete current image wallpaper`: delete the active imported image and fall back to default or a safe previous/default state.
- `Delete unused image wallpapers`: delete imported image files that are not currently active.
- Individual delete from a previously used image item, if multiple retained images are shown.

Avoid a single ambiguous `Remove wallpaper` action unless the label and confirmation make clear whether it only clears the active background or also deletes the underlying imported image file.

## Recommended implementation approach

1. Model chat background state explicitly if it is not already:
   - default / none;
   - solid colour option;
   - image wallpaper reference.
2. Replace or wrap the current picker/import flow so that when an image wallpaper URI is returned:
   - do **not** call `takePersistableUriPermission()` on a `GetContent()` result;
   - open the source URI with `contentResolver.openInputStream(uri)`;
   - decode/validate that it is an image or copy the bytes with sensible size limits;
   - write the imported wallpaper to app-private storage, for example `context.filesDir/chat_wallpapers/<generated-name>.<ext>` or equivalent;
   - store only the app-owned URI/path in `ChatPreferences`.
3. Run image import/copy/decode work off the UI thread, for example in a repository/helper called from the ViewModel on `Dispatchers.IO`. The picker callback should stay lightweight and should not perform large file IO directly in the composable callback.
4. Do not rely on long-term access to the original external provider URI.
5. Defensively handle picker/import failures:
   - catch `SecurityException`, `FileNotFoundException`, `IOException`, and decode failures;
   - show a non-crashing error message/snackbar/toast if import fails;
   - leave the previous background unchanged on failure.
6. Avoid unbounded storage growth:
   - either replace/delete the prior imported wallpaper when a new one is selected, or keep a bounded imported-wallpaper directory with cleanup controls;
   - never leave orphaned app-private wallpaper files with no way for the user to remove them.
7. Add a minimal cleanup UI in Chat Preferences / background settings:
   - expose a clear current image wallpaper delete/remove action where relevant;
   - provide a visible way to delete old/unused imported image wallpapers, such as `Delete unused image wallpapers` or a small `Manage imported images` section;
   - if multiple imported files are retained, allow deleting individual old image wallpapers or deleting all unused image wallpapers;
   - do not delete the currently selected image wallpaper unless the user explicitly chooses to delete it.
8. Ensure clearing an active image wallpaper deletes or safely dereferences the app-owned image where appropriate and resets the preference to default/no-wallpaper state.
9. Keep chat rendering resilient: unreadable/invalid wallpaper references should fail closed and show the normal chat background.
10. Update image preview/rendering code to handle the app-owned reference format chosen by the implementation, whether that is a `file://` URI, absolute private file path, or another app-owned URI representation.

## Legacy / migration behaviour

Existing users may already have `wallpaperType = image` with `wallpaperImageUri` pointing to an external provider `content://` URI.

The implementation must handle this gracefully:

- Do not crash if the stored URI is external, missing, revoked, or unreadable.
- If practical, migrate an existing readable external image URI into app-private storage the next time Chat Preferences is opened or the wallpaper is rendered.
- If migration fails, leave the app usable and fall back to default/no image background with a clear recoverable state.
- Do not delete or mutate unrelated external files.
- Ensure any cleanup action only deletes app-owned files under the wallpaper import directory.

## Storage / retention policy

Prefer a simple bounded policy:

- Store imported images under a dedicated app-private directory, for example `files/chat_wallpapers/`.
- Treat solid colours as preference values, not files.
- Either keep only the current imported image, or keep a small bounded history such as current + a handful of previous images.
- If retaining multiple imported images, determine previous/unused images by scanning the app-owned wallpaper directory and comparing against the active image reference.
- Never auto-delete the active image wallpaper.
- Cleanup UI should only remove app-owned image files, never Gallery/Files source images.

## UI expectations

The cleanup UI should be simple and understandable, not a full wallpaper marketplace.

Acceptable minimal options:

- Background mode choices: `Default`, existing solid colour options, and `Choose image`.
- A `Delete current image wallpaper` action when an imported image is active.
- A `Delete unused image wallpapers` action when old imported image files exist.

If the implementation chooses to support multiple retained image wallpaper files, the UI should make it clear which one is current and allow old ones to be deleted.

## Out of scope

- Redesigning all chat preferences.
- Adding cloud/background sync.
- Building a full wallpaper marketplace.
- Changing the existing solid colour palette unless strictly required for mode modelling.
- Changing chat message rendering, inference, routing, memory, permissions, widgets, or model loading.
- Broad image editing/cropping controls.

## Acceptance criteria

- [ ] Existing solid colour background options still work.
- [ ] Selecting a solid colour does not create image wallpaper files.
- [ ] Selecting Default clears the active custom background mode without crashing.
- [ ] Selecting an image wallpaper from Gallery succeeds.
- [ ] Selecting an image wallpaper from Android Files / file system succeeds without crashing.
- [ ] If an image cannot be opened/decoded/imported, the app shows a graceful error and does not crash.
- [ ] Image import work does not block the main/UI thread.
- [ ] The selected image wallpaper remains available after app restart/process death because it is stored in app-private storage.
- [ ] The app no longer depends on `takePersistableUriPermission()` for chat wallpaper longevity.
- [ ] Existing external `content://` wallpaper preferences are handled gracefully and migrated where practical.
- [ ] Existing wallpaper clear/remove behaviour is preserved or replaced with clearer Default / Delete image actions.
- [ ] Clearing/removing the current image wallpaper removes or safely dereferences the app-owned imported file.
- [ ] Users have an obvious UI path to delete old/unused imported image wallpaper files.
- [ ] Old/unused image wallpaper files do not silently accumulate without cleanup.
- [ ] Image cleanup actions do not delete or affect solid colour options.
- [ ] Image cleanup actions only delete app-owned wallpaper files, never original Gallery/Files source images.
- [ ] Chat rendering remains resilient if the stored image wallpaper reference is missing or unreadable.
- [ ] No inference, routing, memory, permission orchestration, widget, prompt, or model-loading behaviour changes are introduced.

## Validation

Manual smoke on device:

1. Open Chat Preferences / background settings.
2. Select Default and confirm default background is shown.
3. Select each existing solid colour option and confirm the chat background changes correctly.
4. Restart the app and confirm the selected solid colour persists.
5. Select an image from Gallery.
6. Confirm it appears as the chat background.
7. Restart the app and confirm it persists.
8. Select an image using Android Files / file system provider.
9. Confirm the app does not crash and the image wallpaper applies.
10. Restart the app and confirm it persists.
11. Switch from image wallpaper to solid colour and confirm no crash and no stale image overlay.
12. Switch from solid colour back to an imported image if retained.
13. Change image wallpaper multiple times.
14. Use the cleanup UI to delete old/unused image wallpapers.
15. Confirm the currently selected image wallpaper remains unless explicitly deleted.
16. Delete/clear the current image wallpaper and confirm default background returns.
17. Confirm solid colour options still remain available after image cleanup.
18. Try an invalid/non-image file if the picker allows it and confirm graceful failure.
19. Simulate or retain a legacy external `content://` wallpaper preference and confirm the app does not crash.
20. Restart the app after cleanup and confirm there are no crashes or stale missing-wallpaper errors.

lokhor:
## Problem

Selecting a chat background image from Android Files/DocumentsUI crashes the app because `ActivityResultContracts.GetContent()` does not guarantee a persistable URI grant, but `takePersistableUriPermission()` was called unconditionally. Additionally, storing external `content://` URIs in preferences makes the wallpaper dependent on the external provider staying readable across reboots and provider changes.

The wallpaper colour also had a visual cutoff: it was applied only to the message list Column, not the full chat screen, so the composer/input area displayed a separate dark band.

### Image import (new `WallpaperManager` in `:core:memory`)

- `WallpaperManager.importWallpaper(sourceUri)` opens the URI via `ContentResolver.openInputStream()`, validates the bytes are a decodable image, writes a copy to `files/chat_wallpapers/`, and returns the app-private absolute file path
- Import runs on `Dispatchers.IO` — the composable picker callback is lightweight
- Error-handling: `SecurityException`, `FileNotFoundException`, `IOException`, decode failures are all caught; a Snackbar shows the error and the previous background is preserved
- No `takePersistableUriPermission()` call — the app no longer depends on external URI grants for wallpaper longevity

### Wallpaper storage

- `wallpaperImageUri` DataStore key now stores an **app-private absolute file path** (e.g. `/data/data/.../files/chat_wallpapers/wallpaper_xxx.jpg`) instead of an external `content://` URI
- On Chat Preferences open, legacy `content://` wallpaper URIs are automatically migrated to app-private storage. If migration fails, the wallpaper is reset to default (no crash, no stale broken reference)

### Default / Solid / Image mode representation

- The existing three-way `wallpaperType` ("none" / "color" / "image") is unchanged
- Solid colours remain first-class options stored as ARGB longs — never represented as image files
- Default ("none") clears the active custom background without deleting imported image files

### **Fixed: stale `wallpaperImageUri` after mode switch** (#1206 review)

When switching to Default/None or Solid colour, `setWallpaperType()` now **automatically clears `wallpaperImageUri`** (dereferences the active image reference). The imported file is **not deleted** — it becomes eligible for "Delete unused image wallpapers".

- `setWallpaperType("none")` → type = none, image URI cleared
- `setWallpaperType("color")` → type = color, image URI cleared (color picker then sets `wallpaperColor`)
- `setWallpaperType("image")` → type = image, image URI preserved (import path sets URI separately)

### **Fixed: cleanup correctly checks wallpaper mode** (#1206 review)

`deleteUnusedWallpapers()` and `refreshImportedWallpapers()` now only treat the image as **active** when `wallpaperType == "image"`. If the current mode is Default or Solid colour, all imported image files are considered unused and eligible for cleanup.

Rule: `val activeImageUri = wallpaperImageUri.takeIf { wallpaperType == "image" }`

### **Fixed: removed misleading "Previously used images" UI** (#1206 review)

The "Previously used images" section was confusing — tapping items launched the image picker instead of reselecting the displayed image. Replaced with a cleaner, simpler management UI:

- **Choose image** — launches picker
- **Delete current image wallpaper** — visible when mode is Image
- **Delete unused image wallpapers** — visible when unused files exist, shows count

### Background rendering (`ChatScreen.kt`)

- The wallpaper colour background (`Modifier.background(wallpaperBackground)`) was **moved from the inner Column to the root Box** so it covers the full chat screen including the composer/input area
- `wallpaperPainter` now handles both app-private file paths (`BitmapFactory.decodeFile`) and legacy content URIs (`ContentResolver.openInputStream`)
- The InputBar Surface with its tonal elevation sits on top for readability — no abrupt horizontal cutoff

## Validation

- `./gradlew :app:compileDebugKotlin` ✅ (0 errors)
- `./gradlew :feature:settings:test :feature:chat:test` ✅ (all tests pass)

**Required device smoke-testing scenarios:**
1. Start with Default → no image shown
2. Select image from Gallery → image applies and persists
3. Switch to Default → image no longer shown, old image becomes unused
4. Select image again → image applies
5. Switch to Solid colour → colour applies, old image becomes unused
6. Select image from Android Files → no crash, image applies
7. Delete unused image wallpapers → only non-active images deleted
8. Delete current image wallpaper → active image deleted, falls back to Default
9. Solid colours unaffected by image cleanup
10. Restart app → no stale missing-image crashes

Closes #1206

### DIFF
--- a/core/memory/src/main/java/com/kernel/ai/core/memory/wallpaper/WallpaperManager.kt
+++ b/core/memory/src/main/java/com/kernel/ai/core/memory/wallpaper/WallpaperManager.kt
@@ -1,0 +1,149 @@
+package com.kernel.ai.core.memory.wallpaper
+
+import android.content.Context
+import android.graphics.BitmapFactory
+import android.net.Uri
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.File
+import java.io.FileNotFoundException
+import java.io.IOException
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Manages importing, storing, and cleaning up chat wallpaper images in app-private storage.
+ *
+ * Wallpapers are stored under `files/chat_wallpapers/` to decouple wallpaper longevity from
+ * external content-provider URI permissions. The active wallpaper is referenced as an absolute
+ * file path in DataStore preferences.
+ */
+@Singleton
+class WallpaperManager @Inject constructor(
+    @ApplicationContext private val context: Context,
+) {
+    private val wallpaperDir: File
+        get() = File(context.filesDir, WALLPAPER_DIR).also { it.mkdirs() }
+
+    /**
+     * Import a wallpaper from an external [sourceUri] into app-private storage.
+     *
+     * Opens the URI via [ContentResolver.openInputStream], validates the content is a decodable
+     * image, and writes a copy to `files/chat_wallpapers/`.
+     *
+     * @return [Result.success] with the absolute file path, or [Result.failure] with the error.
+     */
+    suspend fun importWallpaper(sourceUri: Uri): Result<String> = withContext(Dispatchers.IO) {
+        try {
+            // Validate the source is a readable image
+            val validateResult = validateImageUri(sourceUri)
+            if (validateResult.isFailure) {
+                return@withContext Result.failure(validateResult.exceptionOrNull()!!)
+            }
+
+            // Read the source bytes
+            val bytes = context.contentResolver.openInputStream(sourceUri)?.use { stream ->
+                stream.readBytes()
+            } ?: return@withContext Result.failure(
+                FileNotFoundException("Cannot open $sourceUri"),
+            )
+
+            // Generate a unique filename
+            val ext = guessExtension(bytes) ?: "jpg"
+            val filename = "wallpaper_${System.currentTimeMillis()}_${sourceUri.hashCode()}.$ext"
+            val destFile = File(wallpaperDir, filename)
+
+            destFile.outputStream().use { output ->
+                output.write(bytes)
+            }
+
+            Result.success(destFile.absolutePath)
+        } catch (e: SecurityException) {
+            Result.failure(e)
+        } catch (e: FileNotFoundException) {
+            Result.failure(e)
+        } catch (e: IOException) {
+            Result.failure(e)
+        }
+    }
+
+    /**
+     * Delete a single wallpaper file identified by its absolute path.
+     * Silently ignores missing files or non-wallpaper-directory paths.
+     */
+    fun deleteWallpaper(filePath: String) {
+        val file = File(filePath)
+        if (file.isFile && file.parentFile?.absolutePath == wallpaperDir.absolutePath) {
+            file.delete()
+        }
+    }
+
+    /**
+     * Delete all imported wallpaper files that are not in [activePaths].
+     *
+     * Only deletes files under the app-owned [wallpaperDir]. Never touches external files.
+     */
+    fun deleteUnusedWallpapers(activePaths: Set<String>) {
+        wallpaperDir.listFiles()?.forEach { file ->
+            if (file.isFile && file.absolutePath !in activePaths) {
+                file.delete()
+            }
+        }
+    }
+
+    /**
+     * Returns all imported wallpaper files sorted by modification time (most recent first).
+     */
+    fun getImportedWallpapers(): List<File> {
+        return wallpaperDir.listFiles()
+            ?.filter { it.isFile }
+            ?.sortedByDescending { it.lastModified() }
+            ?: emptyList()
+    }
+
+    /**
+     * Count of wallpaper files currently in app-private storage.
+     */
+    fun getWallpaperCount(): Int = wallpaperDir.listFiles()?.count { it.isFile } ?: 0
+
+    /**
+     * Validate that [uri] points to a decodable image.
+     */
+    private fun validateImageUri(uri: Uri): Result<Unit> {
+        return try {
+            context.contentResolver.openInputStream(uri)?.use { stream ->
+                val opts = BitmapFactory.Options().apply { inJustDecodeBounds = true }
+                BitmapFactory.decodeStream(stream, null, opts)
+                if (opts.outWidth <= 0 || opts.outHeight <= 0) {
+                    return Result.failure(IllegalArgumentException("URI does not point to a valid image"))
+                }
+            } ?: return Result.failure(FileNotFoundException("Cannot open $uri"))
+            Result.success(Unit)
+        } catch (e: SecurityException) {
+            Result.failure(e)
+        } catch (e: FileNotFoundException) {
+            Result.failure(e)
+        } catch (e: IOException) {
+            Result.failure(e)
+        }
+    }
+
+    /**
+     * Guess a file extension from image magic bytes.
+     */
+    private fun guessExtension(bytes: ByteArray): String? {
+        return when {
+            bytes.size >= 2 && bytes[0] == 0xFF.toByte() && bytes[1] == 0xD8.toByte() -> "jpg"
+            bytes.size >= 8 && bytes[0] == 0x89.toByte() && bytes[1] == 0x50.toByte() &&
+                bytes[2] == 0x4E.toByte() && bytes[3] == 0x47.toByte() -> "png"
+            bytes.size >= 4 && bytes[0] == 0x52.toByte() && bytes[1] == 0x49.toByte() &&
+                bytes[2] == 0x46.toByte() && bytes[3] == 0x46.toByte() -> "webp"
+            else -> null
+        }
+    }
+
+    private companion object {
+        private const val WALLPAPER_DIR = "chat_wallpapers"
+    }
+}
--- a/core/memory/src/main/java/com/kernel/ai/core/memory/wallpaper/WallpaperManager.kt
+++ b/core/memory/src/main/java/com/kernel/ai/core/memory/wallpaper/WallpaperManager.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.withContext
 import java.io.File
 import java.io.FileNotFoundException
 import java.io.IOException
+import java.io.InputStream
+import java.io.OutputStream
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -30,35 +32,58 @@ class WallpaperManager @Inject constructor(
      * Import a wallpaper from an external [sourceUri] into app-private storage.
      *
      * Opens the URI via [ContentResolver.openInputStream], validates the content is a decodable
-     * image, and writes a copy to `files/chat_wallpapers/`.
+     * image, and writes a copy to `files/chat_wallpapers/` using bounded streaming copy.
      *
      * @return [Result.success] with the absolute file path, or [Result.failure] with the error.
      */
     suspend fun importWallpaper(sourceUri: Uri): Result<String> = withContext(Dispatchers.IO) {
         try {
-            // Validate the source is a readable image
+            // Validate the source is a readable image (bounds-only decode — no pixel data loaded)
             val validateResult = validateImageUri(sourceUri)
             if (validateResult.isFailure) {
                 return@withContext Result.failure(validateResult.exceptionOrNull()!!)
             }
 
-            // Read the source bytes
-            val bytes = context.contentResolver.openInputStream(sourceUri)?.use { stream ->
-                stream.readBytes()
-            } ?: return@withContext Result.failure(
-                FileNotFoundException("Cannot open $sourceUri"),
-            )
+            val timestamp = System.currentTimeMillis()
+            val hash = sourceUri.hashCode()
+            val tempFile = File(wallpaperDir, "wallpaper_${timestamp}_${hash}.tmp")
 
-            // Generate a unique filename
-            val ext = guessExtension(bytes) ?: "jpg"
-            val filename = "wallpaper_${System.currentTimeMillis()}_${sourceUri.hashCode()}.$ext"
-            val destFile = File(wallpaperDir, filename)
+            try {
+                // Stream-copy with bounded cap
+                val input = context.contentResolver.openInputStream(sourceUri)
+                    ?: return@withContext Result.failure(
+                        FileNotFoundException("Cannot open $sourceUri"),
+                    )
 
-            destFile.outputStream().use { output ->
-                output.write(bytes)
+                input.use { stream ->
+                    tempFile.outputStream().use { output ->
+                        copyBounded(stream, output, MAX_WALLPAPER_IMPORT_BYTES)
+                    }
+                }
+
+                // Validate the copied file is a fully delivered valid image
+                val fileValidate = validateImageFile(tempFile)
+                if (fileValidate.isFailure) {
+                    tempFile.delete()
+                    return@withContext Result.failure(fileValidate.exceptionOrNull()!!)
+                }
+
+                // Guess extension from magic bytes of the copied file
+                val ext = guessExtensionFromFile(tempFile) ?: "jpg"
+                val destFile = File(wallpaperDir, "wallpaper_${timestamp}_${hash}.$ext")
+                if (!tempFile.renameTo(destFile)) {
+                    // Fallback: temp path used as-is (rare on same filesystem)
+                    return@withContext Result.success(tempFile.absolutePath)
+                }
+
+                Result.success(destFile.absolutePath)
+            } catch (e: Exception) {
+                // Delete any partial file on failure
+                if (tempFile.exists()) tempFile.delete()
+                throw e
             }
-
-            Result.success(destFile.absolutePath)
+        } catch (e: WallpaperImportTooLargeException) {
+            Result.failure(e)
         } catch (e: SecurityException) {
             Result.failure(e)
         } catch (e: FileNotFoundException) {
@@ -109,6 +134,7 @@ class WallpaperManager @Inject constructor(
 
     /**
      * Validate that [uri] points to a decodable image.
+     * Uses bounds-only decoding — no pixel data is loaded into memory.
      */
     private fun validateImageUri(uri: Uri): Result<Unit> {
         return try {
@@ -116,7 +142,9 @@ class WallpaperManager @Inject constructor(
                 val opts = BitmapFactory.Options().apply { inJustDecodeBounds = true }
                 BitmapFactory.decodeStream(stream, null, opts)
                 if (opts.outWidth <= 0 || opts.outHeight <= 0) {
-                    return Result.failure(IllegalArgumentException("URI does not point to a valid image"))
+                    return Result.failure(
+                        IllegalArgumentException("URI does not point to a valid image"),
+                    )
                 }
             } ?: return Result.failure(FileNotFoundException("Cannot open $uri"))
             Result.success(Unit)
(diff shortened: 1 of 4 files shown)


NickMonrad (changes requested):
Thanks — this is directionally good and covers the main Files crash path by removing the unconditional `takePersistableUriPermission()` call and importing selected images into app-private storage. However I found a blocking issue in the image cleanup/state modelling that means the cleanup UI does not work correctly after switching away from an image wallpaper.

### Blocking issue: old image remains treated as active after switching to Default/Solid

The `None` button only calls `setWallpaperType("none")`, and the solid-colour picker sets `wallpaperType = "color"` / `wallpaperColor`, but neither clears `wallpaperImageUri`. That is okay if we deliberately retain the image as history, but the cleanup logic must then treat it as **unused** whenever `wallpaperType != "image"`.

Currently `deleteUnusedWallpapers()` protects `wallpaperImageUri` regardless of the active mode, and `refreshImportedWallpapers()` marks a file active whenever its path equals `wallpaperImageUri`, again regardless of whether the active mode is `image`. Result: after selecting Default or a solid colour, the previously selected image is still marked active internally, `Delete unused image wallpapers` may not appear, and a single old imported image can become impossible to delete from the UI unless the user reactivates image mode first.

Please change the active-path / active-indicator logic to include the current image only when `wallpaperType.first() == "image"`. For example:

```kotlin
val type = chatPreferences.wallpaperType.first()
val currentUri = chatPreferences.wallpaperImageUri.first()
val active = if (type == "image" && currentUri != null) setOf(currentUri) else emptySet()
```

and apply the same rule in `refreshImportedWallpapers()` so `isActive` only true when the active mode is image.

### Also fix or remove the broken Previous images click behaviour

The `Previously used images` rows currently attach a click handler to non-active rows that launches the system image picker again. That is surprising and does not reselect or delete the displayed wallpaper. Either wire this to reselect the existing app-owned wallpaper path, add an explicit delete action, or remove the click affordance entirely and rely on the Delete unused action.

### Non-blocking but worth considering

`WallpaperManager.importWallpaper()` validates the image, then reads the whole source into memory with `readBytes()`. This is better than the old permission crash, but a very large file selected from Files could still put pressure on memory. A streaming copy with a max-byte guard would be more robust.

CI was still in progress when I reviewed. I’d re-run/confirm CI after the fixes and smoke test: image → Default → Delete unused; image → Solid colour → Delete unused; and multiple imported images with one active vs none active.

lokhor:
## Review fixes pushed

Pushed `2ef7af8c` with fixes for all three blocking issues:

### 1. Stale `wallpaperImageUri` after mode switch
`setWallpaperType()` now **clears `wallpaperImageUri`** when switching to non-image modes (`"none"`, `"color"`). The imported file is **not deleted** — it becomes eligible for "Delete unused image wallpapers".

### 2. Cleanup now respects wallpaper mode
`deleteUnusedWallpapers()` and `refreshImportedWallpapers()` both check `wallpaperType == "image"` before treating the current URI as active. If mode is Default or Solid colour, **all** imported images are unused.

### 3. Removed confusing "Previously used images" UI
Opened for Option B (simplify). The wallpaper management section now shows only:
- **Choose image** — launches picker
- **Delete current image wallpaper** — when type is Image
- **Delete unused image wallpapers** — when unused files exist

---

Build: `./gradlew :app:compileDebugKotlin` ✅  
Tests: `./gradlew :feature:settings:test :feature:chat:test` ✅

Ready for device smoke testing.

NickMonrad (approved):
Re-reviewed PR #1208 after the requested fixes.

The previous blocking feedback has been addressed:

- Switching away from image mode now clears/dereferences `wallpaperImageUri` in `setWallpaperType()` when the new mode is not `image`, so old imported images become eligible for unused cleanup rather than being treated as active.
- Cleanup now only preserves the active image when `wallpaperType == "image"`; Default/Solid mode correctly treats imported files as unused.
- The misleading “Previously used images” UI has been removed in favour of a simpler management flow: Choose image, Delete current image wallpaper, and Delete unused image wallpapers.
- The picker callback now delegates to the ViewModel/import path and no longer calls `takePersistableUriPermission()` directly.
- The wallpaper/background rendering change now applies the solid background at the root chat content box, so the composer area sits on top of the active background instead of creating the earlier hard cutoff.

CI is green on the latest head. I did not find a remaining blocker in the diff.

Non-blocking future hardening: `WallpaperManager.importWallpaper()` currently reads the full source image into memory with `readBytes()` after validation. This is probably fine for normal Gallery/Files images, but a future polish pass could stream-copy with a size cap to avoid pathological huge-file memory pressure.

Approved, pending the manual device smoke from #1206, especially Files picker selection, image → default/solid cleanup, delete-current/delete-unused actions, restart persistence, and keyboard-open background continuity.

NickMonrad:
One more safety hardening item before merge would be worthwhile.

The current `WallpaperManager.importWallpaper()` validates the URI and then reads the whole selected image into memory with `stream.readBytes()` before writing it to app-private storage. That is much safer than the old persistable-permission path, but it can still create avoidable memory pressure if a user picks a very large image file from Files.

Please update the import path to stream-copy the image into the destination file with a bounded maximum size instead of loading the full file into a `ByteArray`.

Suggested approach:

- Keep decode-bounds validation using `BitmapFactory.Options(inJustDecodeBounds = true)`.
- Decide a reasonable maximum source/import size, for example 15–25 MB. Use a named constant.
- Open a fresh input stream for copying after validation.
- Copy using a fixed buffer, counting bytes as you go.
- If the byte count exceeds the max:
  - stop copying;
  - delete the partial destination file;
  - return `Result.failure(IllegalArgumentException("Wallpaper image is too large"))` or similar.
- Avoid `readBytes()`.
- Guess the extension without requiring the full file in memory. Options:
  - use `contentResolver.getType(uri)` and map `image/jpeg`, `image/png`, `image/webp`; or
  - read only the first few header bytes from a short stream.
- Keep existing failure handling for `SecurityException`, `FileNotFoundException`, `IOException`, and decode failures.

This should keep #1206 robust for pathological/huge Files selections without changing the UX or broader wallpaper state model.

lokhor:
## Fixed: bounded streaming copy for wallpaper import (`60fd93e6`)

Replaced `stream.readBytes()` (loaded entire file into memory) with bounded streaming copy.

**Changes:**
- `copyBounded()` — 8 KB buffer streaming copy with a **25 MB maximum import size** cap
- `validateImageFile()` — post-copy bounds-only decode of the written file (no pixel data loaded) to verify full delivery
- `guessExtensionFromFile()` — reads only the first 16 bytes from the written file for magic-byte extension detection (jpg/png/webp)
- **Partial file cleanup** — on any failure (too large, decode failure, I/O error), the `.tmp` file is deleted
- All existing error paths preserved (`SecurityException`, `FileNotFoundException`, `IOException`)
- `WallpaperImportTooLargeException` extends `IOException` for consistent propagation

**No regressions in previously fixed behaviour:**
- `wallpaperImageUri` cleared on Default/Solid colour switch ✅
- Cleanup respects `wallpaperType == "image"` ✅
- Background extends behind composer ✅
- Gallery and Android Files selection both work ✅

**Validation:**
- `./gradlew :app:compileDebugKotlin` ✅
- `./gradlew :core:memory:testDebugUnitTest :feature:settings:testDebugUnitTest` ✅

NickMonrad (approved):
Re-reviewed latest PR #1208 after the bounded streaming-copy hardening.

The follow-up is implemented correctly:

- `WallpaperManager.importWallpaper()` no longer uses whole-file `readBytes()`; it validates the source with bounds-only decode, then stream-copies into a temp file with a fixed buffer and a named 25 MB cap.
- Partial temp files are deleted on failure, including oversized imports.
- The copied file is validated before activation and renamed using detected magic-byte extension where possible.
- Import still runs on `Dispatchers.IO` and preserves the previous Result-based error path for the UI.
- Prior fixes remain intact: switching away from image mode clears the active image reference, cleanup only treats an image as active when `wallpaperType == "image"`, the misleading previously-used UI is gone, and chat background rendering still applies at the root behind the composer.
- CI is green.

Approved. Device smoke should still be performed before merge, especially Gallery import, Files import, oversized/non-image failure if practical, switch image → default/solid, delete unused/current wallpaper, and keyboard-open composer background continuity.

NickMonrad (changes requested):
New device-smoke UX issue from PR #1208:

The current UI shows `Delete unused image wallpapers (2)`, which confirms imported app-owned wallpapers are being retained, but there is no way to reselect those already-imported/on-device wallpapers. The only available image action is `Image`, which launches the external picker again.

That creates a confusing/unsafe wallpaper-management flow:

- User imports image A.
- User switches to Default/Solid or imports image B.
- Image A remains in app-private storage and is counted as unused.
- The user can delete A, but cannot choose A again without finding it in Gallery/Files and re-importing it.

Because PR #1208 retains previous imported wallpapers and exposes their count, we should either:

1. Add a small `Imported wallpapers` / `Saved wallpapers` section that lists app-owned images and allows reselecting them, with a current indicator and delete action for unused items; or
2. Change the retention model to keep only the current image and delete/dereference the previous app-owned image when switching away, so the UI does not imply saved wallpapers are available for management.

Option 1 is the better UX and aligns with the earlier #1206 discussion: if imported wallpapers exist on-device, the user should be able to choose one without going back through Gallery/Files.

Suggested minimal fix:

- Show a `Saved image wallpapers` row/grid/list when `importedWallpapers` is non-empty.
- Each item should show a thumbnail or filename/preview.
- Tapping an item sets `wallpaperType = "image"` and `wallpaperImageUri = item.path`.
- The active item should be clearly marked.
- Keep `Delete unused image wallpapers` for bulk cleanup.
- Keep `Delete current image wallpaper` for explicit active deletion.
- Solid colour and Default modes should remain first-class and unaffected.

Requesting changes so this is fixed before merge rather than leaving the retained wallpapers as delete-only orphaned state.

lokhor:
## Added: Saved image wallpapers section (`d0bd425b`)

Imported wallpapers can now be reselected from a new **"Saved image wallpapers"** section in Chat Preferences without launching Gallery/Files again.

### UX
- **Thumbnails in a horizontally scrollable row** — 80dp rounded squares using `AsyncImage` with `Uri.fromFile(File(path))`
- **Active indicator** — checkmark overlay + primary-toned border when `wallpaperType == "image"` and the item is the current `wallpaperImageUri`
- **Tap to select** — sets `wallpaperType = "image"` and `wallpaperImageUri = path` without re-importing
- **Missing file handling** — `selectImportedWallpaper` checks `File(path).exists()` before activating; refreshes the list and shows an error snackbar if the file was deleted
- **Existing controls preserved** — "Delete current image wallpaper" (visible when type is Image) and "Delete unused image wallpapers (N)" both unchanged

### Behaviour
- If `wallpaperType != "image"`, no thumbnail appears active (the existing `isActive` logic in `refreshImportedWallpapers` uses `takeIf { currentType == "image" }`)
- Saved wallpapers remain listed after switching to Default/Solid — eligible for reselection
- Bulk unused cleanup never deletes the active image
- Section hides when no imported wallpapers exist

### Validation
- `./gradlew :app:compileDebugKotlin` ✅
- `./gradlew :core:memory:testDebugUnitTest :feature:settings:testDebugUnitTest` ✅

**Ready for device smoke testing per the 20-step checklist.**

NickMonrad (approved):
Re-reviewed latest PR #1208 after the saved-wallpapers UX fix and device smoke confirmation.

The previous UX blocker is resolved:

- `ChatPreferencesViewModel` now exposes `selectImportedWallpaper(path)`, verifies the file exists, then sets `wallpaperImageUri` and `wallpaperType = "image"` without re-importing or launching Gallery/Files.
- `ChatPreferencesScreen` now shows a `Saved image wallpapers` section when imported wallpapers exist.
- Saved wallpapers render as tappable thumbnails using app-owned file references.
- The active saved wallpaper has a visible selected border/check indicator.
- Bulk unused cleanup and explicit current-image deletion remain available.

The earlier robustness changes also remain in place: bounded streaming import with a 25 MB cap, no `takePersistableUriPermission()`, legacy migration handling, image/default/solid mode semantics, cleanup scoped to app-owned files, and root-level chat background rendering behind the composer.

CI is green, and the user confirmed device testing works well.

Approved.

Issue #1214, opened 2026-06-13
Bug: Learn example for dentist appointment falls through to LLM


NickMonrad:
## Summary

One of the **Learn what Jandal can do** example actions is being routed to generic LLM fallthrough instead of a deterministic app action / supported intent path.

## Observed behaviour

From device testing:

```text
Heard: Add dentist appointment next Tuesday
Action: llm_fallthrough
Result: Sending to Jandal for processing…
```

## Expected behaviour

Selecting or speaking this Learn example should not fall through to the LLM.

Expected handling should be one of:

1. Route to the appropriate calendar/reminder intent if appointments are currently supported; or
2. Use slot filling / clarification if the appointment action requires missing fields such as time; or
3. Change/remove the Learn example if Jandal does not yet support this capability reliably.

The Learn catalogue should only advertise prompts that map to supported deterministic behaviours or safe prefill flows.

## Why this matters

The Learn surface is a user trust and discoverability path. If an advertised action falls through to the LLM, users may assume the app can perform an action that is not actually wired to a local capability.

## Likely area

- Learn what Jandal can do example catalogue / prompt metadata
- Quick Intent Router matching for appointment/calendar/reminder phrasing
- Action prefill path from Learn examples into Actions

Related context:

- #751 navigation/tool discoverability epic
- #1113 test harness / QIR evidence epic
- #1211 Tools and Learn examples search/filtering should preserve the rule that Learn examples are safe and supported

## Acceptance criteria

- [ ] Reproduce the current failure with `Add dentist appointment next Tuesday` and capture the actual route.
- [ ] Decide whether this example should be backed by calendar/reminder intent support or removed/reworded from Learn.
- [ ] If supported, update QIR/example metadata so the utterance does not return `llm_fallthrough`.
- [ ] If extra details are required, route into slot filling rather than LLM fallthrough.
- [ ] Add/update a regression test for the Learn example phrase.
- [ ] Include evidence showing the route is no longer `llm_fallthrough`.

## Suggested test cases

```text
Add dentist appointment next Tuesday
Add a dentist appointment next Tuesday
Schedule dentist appointment next Tuesday
Remind me about my dentist appointment next Tuesday
```

The exact expected intent may vary depending on whether appointment/calendar support is implemented, but none of the advertised Learn examples should silently fall through to LLM processing.

NickMonrad:
Fixed by merged PR #1216.

[...]

PR #1216, opened 2026-06-13
feat(#1215): add Learn example test harness with QIR regression coverage

### LINKED ISSUE #1215: Child issue: Add test harness coverage for every Learn example action
# Child issue: Add test harness coverage for every Learn example action

Parent epic: #751 — Launch navigation and tool discoverability refactor

Evidence alignment: #1113 — GitHub-native test evidence dashboard for CI and on-device results

Related:

- #1211 — Add Tools search and filtering for destinations and examples
- #1214 — Bug: Learn example for dentist appointment falls through to LLM

## Summary

Extend the test harness so every action/example surfaced by **Learn what Jandal can do** is covered by automated validation.

The immediate trigger is #1214, where an advertised Learn example falls through to `llm_fallthrough`:

```text
Heard: Add dentist appointment next Tuesday
Action: llm_fallthrough
Result: Sending to Jandal for processing…
```

That should not be discovered one example at a time during manual testing. The catalogue and the router should have a regression safety net.

## Problem

The Learn surface is now a user-facing contract: it tells users what Jandal can do.

If a Learn example:

- routes to `llm_fallthrough`,
- depends on unsupported capability,
- has stale prompt wording,
- opens the wrong screen,
- executes automatically when it should only prefill,
- or requires slot filling but bypasses it,

then the app is advertising behaviour that is not actually reliable.

## UX / product rule

Every Learn example must be one of:

1. **Deterministic supported action** — routes to a known app action/tool/intent.
2. **Supported slot-fill action** — routes to a known slot-fill flow when required fields are missing.
3. **Safe prefill only** — opens Actions with the example text populated, without automatic execution.
4. **Explicitly non-executable educational item** — only if labelled as such and not presented as something Jandal will do immediately.

No Learn example should silently fall through to generic LLM processing unless it is intentionally labelled as a free-form chat/example prompt.

### 1. Make Learn examples testable from source of truth

Avoid duplicating a static list of examples in the test harness if possible.

Prefer a shared/local catalogue model that exposes enough metadata for each Learn example:

```kotlin
data class LearnExample(
    val id: String,
    val title: String,
    val prompt: String,
    val category: String,
    val expectedMode: ExpectedLearnMode,
    val expectedRoute: String? = null,
    val requiresSlotFill: Boolean = false,
    val prefillOnly: Boolean = false,
)
```

Suggested expected modes:

```text
intent_route
slot_fill
prefill_only
non_executable_info
freeform_chat_allowed
```

Only use `freeform_chat_allowed` where the catalogue deliberately presents the item as open-ended chat, not as an app action.

### 2. Add catalogue integrity tests

Add host-side/unit tests that assert:

- every Learn example has a stable `id`;
- every example has non-empty title/prompt/category;
- every executable example declares an expected route or expected slot-fill behaviour;
- no executable example uses `llm_fallthrough` as its expected path;
- duplicate prompts/ids are rejected;
- examples added later automatically fail tests unless metadata is complete.

### 3. Add QIR / router regression tests for every executable prompt

For each Learn example with `expectedMode = intent_route` or `slot_fill`:

- feed the prompt text into the same Quick Intent Router path used by Actions/voice where practical;
- assert the route is not `llm_fallthrough`;
- assert the route matches the declared expected route, or that the expected slot-fill flow is triggered;
- record the result per example id.

### 4. Add UI harness coverage for Learn selection behaviour

For Learn examples that should open Actions as a draft/prefill:

- open **Learn what Jandal can do**;
- select the example;
- assert Actions opens with the expected text populated;
- assert the example is not automatically executed;
- assert navigation back behaviour remains correct.

This does not need to run every heavy LLM/tool action on device. The critical invariant is that the Learn surface points to the correct deterministic/prefill path.

### 5. Add test report evidence

Extend the existing test report schema/dashboard to include a Learn catalogue suite, for example:

```json
{
  "suite": "learn_examples",
  "summary": {
    "total": 24,
    "passed": 24,
    "failed": 0
  },
  "cases": [
    {
      "name": "learn_add_dentist_appointment_next_tuesday",
      "learn_example_id": "calendar_dentist_next_tuesday",
      "prompt": "Add dentist appointment next Tuesday",
      "expected_mode": "slot_fill",
      "expected_route": "calendar_create_event",
      "actual_route": "calendar_create_event",
      "passed": true,
      "failure_category": null
    }
  ]
}
```

Suggested failure categories:

```text
missing_catalogue_metadata
unsupported_learn_example
llm_fallthrough
wrong_route
missing_slot_fill
unexpected_execution
prefill_mismatch
navigation_mismatch
```

## Device strategy

Use a layered approach so the suite is useful without becoming slow/flaky:

- **CI/host tests:** catalogue integrity and router expectations where dependencies allow.
- **S21 on-device:** primary UI smoke path for Learn selection/prefill behaviour.
- **S23 Ultra on-device:** targeted confirmation only if model/runtime-specific behaviour is relevant.

The suite should not require running every example through an actual local LLM model unless the example is explicitly meant to exercise model/tool generation.

## Acceptance criteria

- [ ] The Learn catalogue has a testable source of truth or metadata layer.
- [ ] Every Learn example has a stable id, category, prompt, and expected behaviour.
- [ ] Automated tests enumerate all Learn examples from the catalogue rather than relying on a manually maintained duplicate list.
- [ ] Executable Learn examples fail tests if they route to `llm_fallthrough`.
- [ ] Slot-fill examples fail tests if the expected slot-fill path is not triggered.
- [ ] Prefill-only examples fail tests if they execute automatically or populate the wrong text.
- [ ] `Add dentist appointment next Tuesday` is included as a regression case.
- [ ] Test evidence is published under a `learn_examples` suite in the existing report/dashboard flow.
- [ ] Documentation explains how to add a new Learn example and its corresponding expected test metadata.

## Suggested initial test cases

Start by enumerating the actual catalogue, then make sure at minimum the current failing example is represented:

```text
Add dentist appointment next Tuesday
```

The final implementation should not stop at a hand-written subset; it should cover every current and future Learn example automatically through catalogue enumeration.

lokhor:
## Summary

Implement automated coverage for every action/example surfaced by **Learn what Jandal can do**, treating the Learn catalogue as a user-facing contract.

### Data model & catalogue refactor
- Introduced `LearnExample` data model with `ExpectedLearnMode` enum (QirIntent, QirSlotFill, MealPlannerHandoff, FreeformChatAllowed, PrefillOnly, NonExecutableInfo)
- Added `expectedMissingSlot` field for QirSlotFill examples
- Refactored `ToolsLearnScreen.kt` to derive UI from shared model — one source of truth
- Added "Plan meals" and "Plan dinners for this week" as meal-planner handoff prompts (67 total)

### QIR fixes (10+ pattern gaps)
Detailed in evidence: 10/10 on-device validated fixes for weather patterns, calendar regex, email routing, etc.

### Calendar regex hardening
- Restructured bare-noun calendar pattern to require **event-word OR date reference** after the noun
- Food/drink nouns (dinner, lunch, coffee, drinks) only match as calendar if followed by a date context
- Added negative regression tests ensuring list commands are not stolen

### QirSlotFill classification
- `calendar_soccer_training` → `QirSlotFill` (missing `date`)
- `people_email_alex` → `QirSlotFill` (missing `body`)
- `people_email_school` → `QirSlotFill` (missing `body`)
- Routing test now asserts `NeedsSlot` + correct `missingSlot.name`
- QirIntent test now fails if result is `NeedsSlot` (catches misclassification)
- Catalogue integrity validates `expectedMissingSlot` on QirSlotFill examples

### Tests
- **New: `LearnExampleRoutingTest`** — catalogue-driven QIR routing test that enumerates the real `allLearnExamples` list and fails on wrong routes
- **Catalogue integrity tests**: metadata completeness, unique ids/prompts, expectedRoute/expectedMissingSlot validation
- **QIR coverage report** trimmed: removed duplicated Learn-specific phrase lists
- **Negative calendar regression tests**: 5 new tests in `QuickIntentRouterListRoutingTest`
- All unit tests pass: 1556 core skills + 701 app tests
- CI green

### Evidence
- Published to `test-results` branch: `results/pr/1216/on_device/__s21-exynos_learn_examples.json`
- Local evidence file: scope `targeted_qir_regression_subset`, 14 cases

## Acceptance criteria
- [x] Learn catalogue has testable source of truth
- [x] Every Learn example has stable id, category, prompt, expected mode, expected route
- [x] Tests enumerate all examples from the real catalogue
- [x] Deterministic examples fail if they route to `llm_fallthrough`
- [x] QirIntent examples fail if they unexpectedly return NeedsSlot
- [x] QirSlotFill examples assert NeedsSlot + correct missingSlot name
- [x] Meal-planner examples fail if not `start_meal_planner`
- [x] Free-form examples explicitly marked as freeform_chat_allowed
- [x] Broad calendar regex no longer steals list commands
- [x] Existing Android UI test tags preserved (`tools_learn_` prefix)
- [x] #1214 remains fixed
- [x] S21 validation passed (10/10 QIR patterns)
- [x] Evidence published (accurate scope)
- [x] CI unit tests pass (core:skills + :app)

Closes #1215

### DIFF
--- a/app/src/main/java/com/kernel/ai/navigation/LearnExample.kt
+++ b/app/src/main/java/com/kernel/ai/navigation/LearnExample.kt
@@ -55,10 +55,11 @@ enum class ExpectedLearnMode {
  * @property title User-visible label shown on the Learn screen.
  * @property prompt The text sent to the app when the example is selected.
  *   Defaults to [title] when not specified.
- * @property category Section/category name.
- * @property expectedMode How this example should be routed (see [ExpectedLearnMode]).
  * @property expectedRoute Expected QIR intent name (e.g. "create_calendar_event").
  *   Required for [QirIntent] and [QirSlotFill]; null for others.
+ * @property expectedMissingSlot For [QirSlotFill] examples, the name of the slot
+ *   the router expects the user to fill (e.g. "date", "body").
+ *   Required for [QirSlotFill]; null for others.
  * @property prefillOnly When true, selecting this example opens Actions as a draft
  *   without auto-execution. Defaults to true for safety.
  */
@@ -69,6 +70,7 @@ data class LearnExample(
     val category: String,
     val expectedMode: ExpectedLearnMode,
     val expectedRoute: String? = null,
+    val expectedMissingSlot: String? = null,
     val prefillOnly: Boolean = true,
 )
 
@@ -136,7 +138,8 @@ val allLearnExamples: List<LearnExample> = listOf(
     LearnExample("time_timer_10", "Set a timer for 10 minutes", category = "Time & planning",
         expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "set_timer"),
     LearnExample("calendar_soccer_training", "Create a calendar event for soccer training", category = "Time & planning",
-        expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "create_calendar_event"),
+        expectedMode = ExpectedLearnMode.QirSlotFill, expectedRoute = "create_calendar_event",
+        expectedMissingSlot = "date"),
     LearnExample("calendar_dentist", "Add dentist appointment next Tuesday", category = "Time & planning",
         expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "create_calendar_event"),
     LearnExample("calendar_assembly", "Schedule school assembly for Friday", category = "Time & planning",
@@ -170,13 +173,14 @@ val allLearnExamples: List<LearnExample> = listOf(
     LearnExample("weather_hot_tomorrow", "How hot will it be tomorrow?", category = "Weather",
         expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "get_weather"),
 
-    // ── People & communication ─────────────────────────────────────────────────
     LearnExample("people_email_alex", "Email Alex about the meeting", category = "People & communication",
-        expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "send_email"),
+        expectedMode = ExpectedLearnMode.QirSlotFill, expectedRoute = "send_email",
+        expectedMissingSlot = "body"),
     LearnExample("people_text_running_late", "Text Sarah that I'm running late", category = "People & communication",
         expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "send_sms"),
     LearnExample("people_email_school", "Email school that Freyja is sick today", category = "People & communication",
-        expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "send_email"),
+        expectedMode = ExpectedLearnMode.QirSlotFill, expectedRoute = "send_email",
+        expectedMissingSlot = "body"),
     LearnExample("people_text_dad", "Text Dad that I'm on my way", category = "People & communication",
         expectedMode = ExpectedLearnMode.QirIntent, expectedRoute = "send_sms"),
     LearnExample("people_call_mum", "Call Mum", category = "People & communication",
--- a/app/src/main/java/com/kernel/ai/navigation/LearnExample.kt
+++ b/app/src/main/java/com/kernel/ai/navigation/LearnExample.kt
@@ -1,0 +1,242 @@
+package com.kernel.ai.navigation
+
+/**
+ * Expected routing mode for a [LearnExample] in the "Learn what Jandal can do" catalogue.
+ *
+ * Every Learn example must declare one of these modes so the test harness can assert
+ * correct routing behaviour and detect regressions.
+ */
+enum class ExpectedLearnMode {
+    /**
+     * Routes through the Quick Intent Router deterministic regex/classifier.
+     * Must NOT produce "llm_fallthrough".
+     */
+    QirIntent,
+
+    /**
+     * Routes through QIR but triggers a slot-fill flow because required fields
+     * are missing from the prompt (e.g. "Create a calendar event" without a title).
+     */
+    QirSlotFill,
+
+    /**
+     * Intentional handoff to the meal planner via QIR's "start_meal_planner" route.
+     * Must NOT be generic "llm_fallthrough".
+     */
+    MealPlannerHandoff,
+
+    /**
+     * Deliberately open-ended chat example outside the deterministic action scope.
+     * May produce "llm_fallthrough" but only because the metadata explicitly allows it.
+     */
+    FreeformChatAllowed,
+
+    /**
+     * Opens Actions with the prompt text pre-filled but does NOT auto-execute.
+     * Examples that require user context or confirmation before execution.
+     */
+    PrefillOnly,
+
+    /**
+     * Non-executable educational item — displayed as informational content only.
+     * Not expected to route or trigger any action.
+     */
+    NonExecutableInfo,
+}
+
+/**
+ * A single example in the "Learn what Jandal can do" catalogue.
+ *
+ * This model is the shared source of truth for both the UI and the test harness.
+ * Tests enumerate the actual [allLearnExamples] list rather than maintaining a
+ * separate hard-coded list that can drift.
+ *
+ * @property id Stable unique identifier (used for test tags and evidence keys).
+ * @property title User-visible label shown on the Learn screen.
+ * @property prompt The text sent to the app when the example is selected.
+ *   Defaults to [title] when not specified.
+ * @property category Section/category name.
+ * @property expectedMode How this example should be routed (see [ExpectedLearnMode]).
+ * @property expectedRoute Expected QIR intent name (e.g. "create_calendar_event").
+ *   Required for [QirIntent] and [QirSlotFill]; null for others.
+ * @property prefillOnly When true, selecting this example opens Actions as a draft
+ *   without auto-execution. Defaults to true for safety.
+ */
+data class LearnExample(
+    val id: String,
+    val title: String,
+    val prompt: String = title,
+    val category: String,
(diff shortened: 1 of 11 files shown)


NickMonrad (changes requested):
Thanks — this is directionally the right shape for #1215: the catalogue model is a good start, the missing `Plan meals` prompt is present, and CI is green on the PR head.

I’m requesting changes because there are a few correctness issues that undermine the main goal of the issue: reliable Learn coverage from the real catalogue.

Blocking items:

1. **Existing Learn UI Android tests are now broken.** `ToolsLearnScreen` now uses raw catalogue ids as test tags, but `ToolsHubNavTest` still looks for the existing `tools_learn_*` tags. CI did not catch this because those Android tests are not part of the PR CI run. Keep the existing public test tags, or update all affected Android tests/docs deliberately. Example mismatch: `ToolsLearnScreen.kt` emits `lists_add_milk`, while `ToolsHubNavTest.kt` still searches for `tools_learn_lists_add_milk`.

2. **The router coverage does not actually enumerate the real Learn catalogue and does not fail on wrong routes.** The new catalogue integrity test enumerates `allLearnExamples`, but the QIR regression coverage adds another hard-coded set of `learnExample*Phrases()` lists. Worse, the existing coverage report explicitly does not fail when routes are wrong. That means a future Learn prompt can still drift away from QIR while CI stays green, which is the core failure #1215 is meant to prevent. There is also a concrete mismatch: `learnExampleCalculationPhrases()` includes arithmetic, unit conversion, and cooking conversion prompts but is registered with expected intent `convert_currency`.

3. **The new bare calendar regex is too broad and unanchored.** Because it includes nouns like `coffee`, `dinner`, and `lunch`, has an optional date suffix, and is tried before list patterns, it can intercept non-calendar commands such as `add coffee to my shopping list` or `put dinner on my list` as `create_calendar_event`. It needs to require a real calendar/event boundary or date, be anchored appropriately, and include negative regression tests for list commands using these nouns.

4. **The checked-in learn evidence claims 67/67 passed but only includes 14 cases.** `issue-1215-learn-examples-evidence.json` has `summary.total = 67` and `passed = 67`, but the `cases` array only contains 14 entries. That makes the evidence misleading. Either include all 67 cases or change the summary to the actual subset and label it as targeted evidence.

Once those are fixed, I’d re-review. The shape is close, but these need to be tightened before merge.

NickMonrad (approved):
Re-reviewed the updated PR. The previous blockers are addressed:

- Learn UI test tags now preserve the existing `tools_learn_` prefix.
- `LearnExampleRoutingTest` now enumerates the real `allLearnExamples` catalogue and fails deterministic examples on fallthrough/wrong route.
- Learn-specific hard-coded coverage was removed from the diagnostic QIR coverage report path.
- Calendar/list stealing is covered by explicit negative list-routing tests.
- The checked-in evidence summary now correctly declares a targeted 14-case subset instead of claiming full 67/67 on-device evidence.
- CI is green on the updated head.

One non-blocking follow-up to consider later: the catalogue currently uses `QirIntent` for some prompts that likely result in `NeedsSlot` at runtime, such as calendar/email examples with missing details. The route-level protection is now in place, so I’m comfortable approving this PR, but a future cleanup could use `QirSlotFill` more precisely and assert the actual `NeedsSlot` result type for those examples.

NickMonrad (changes requested):
I’m changing my previous non-blocking note into a before-merge change.

Reason: #1215 is specifically about the Learn catalogue becoming a testable contract. If examples that actually enter slot-fill are marked as plain `QirIntent`, the tests only prove the broad route name and don’t prove the user-visible behaviour that matters: the expected follow-up/slot-fill flow.

Please tighten this now in the same PR:

1. Reclassify Learn examples that return `QuickIntentRouter.RouteResult.NeedsSlot` as `ExpectedLearnMode.QirSlotFill`, not `QirIntent`.
   - Likely candidates include calendar/email/phone examples where required fields are missing or partially supplied, for example:
     - `Create a calendar event for soccer training`
     - `Add dentist appointment next Tuesday`
     - `Schedule school assembly for Friday`
     - `Email Alex about the meeting`
     - any other catalogue prompt that the current router returns as `NeedsSlot`.

2. Update `LearnExampleRoutingTest` so `QirSlotFill` examples assert the actual result type is `NeedsSlot`, not merely “some route with the right intent”. The test should also assert the intent name matches `expectedRoute` and, where practical, the missing slot name is the expected one.

3. Add metadata if needed, for example `expectedMissingSlot: String?`, so the catalogue can declare the expected follow-up field for slot-fill examples.

4. Keep `QirIntent` for examples that are fully actionable without slot-fill.

5. Update the Learn examples guide and targeted evidence if the expected mode values change there.

This should make the harness more valuable: it will catch both raw fallthrough regressions and accidental loss of the expected slot-fill flow.

NickMonrad (approved):
Re-reviewed the latest update after the slot-fill tightening. This now addresses the previous before-merge request:

- `LearnExample` includes `expectedMissingSlot`.
- Known slot-fill examples are classified as `QirSlotFill` with expected missing slots.
- `LearnExampleRoutingTest` now fails if a `QirIntent` unexpectedly returns `NeedsSlot`.
- `QirSlotFill` examples now require an actual `NeedsSlot` result and assert the missing slot name.
- Catalogue integrity tests enforce `expectedMissingSlot` for slot-fill examples and reject it for plain `QirIntent` examples.
- The guide now documents the distinction between `QirIntent` and `QirSlotFill`.
- CI is green on the latest head.

One small non-blocking doc nit: `docs/agents/learn-examples-guide.md` says the Learn examples in `QuickIntentRouterTest` must pass, but the catalogue-driven routing test is now `LearnExampleRoutingTest`. That can be cleaned up in this PR or later; it does not affect correctness.

Approved.

Issue #1295, opened 2026-06-19
Publish raw harness JSON and dashboard evidence for launch validation


NickMonrad:
Parent/related: #1113, #1287, #427
Evidence source: PR #1292

## Problem

PR #1292 publishes a durable Markdown summary for the S21 launch validation run, but explicitly states that raw harness JSON was not committed and dashboard publication was not performed.

#1287 requires launch validation evidence to be committed or linked and published clearly, including separate S21 and S23U evidence.

## Scope

Close the evidence-publication gap for launch validation runs.

This should cover:

- Raw harness JSON artifact capture.
- Durable repo storage or durable artifact linking for S21 launch validation runs.
- Dashboard ingestion/publication for S21 full launch-scope evidence.
- Dashboard ingestion/publication for S23U targeted smoke/comparison evidence once available.
- Clear dashboard labelling so S23U targeted smoke is not misrepresented as full-suite coverage.
- A repeatable command or workflow for future PR/device evidence publication.

## Desired behaviour

For launch validation, a reviewer should be able to see:

- device model and serial/connection type;
- app build/commit SHA;
- timestamp;
- raw JSON evidence or durable artifact link;
- Markdown summary;
- dashboard entry;
- pass/fail/xfail/skipped/not-reached/excluded totals;
- classification of launch-blocking, launch-deferred, fixture/harness limitation, and accepted known limitations.

## Acceptance criteria

- Identify the current raw JSON output path(s) for the S21 run referenced by PR #1292, or document why they are no longer available.
- Commit raw JSON evidence under the repo evidence path or add durable links if committing raw artifacts is not appropriate.
- Publish S21 launch validation evidence to the dashboard or document the currently missing dashboard capability.
- Add/extend automation so future harness runs can publish raw JSON + Markdown + dashboard evidence with minimal manual work.
- Ensure #1287 links to the final S21 and S23U evidence locations.
- Dashboard clearly distinguishes full S21 launch-scope evidence from targeted S23U daily-driver smoke.

## Launch posture

This is part of making launch readiness auditable. It does not necessarily block merging PR #1292 if the Markdown summary is accepted, but #1287 should remain open until raw/dashboard evidence and S23U targeted smoke are complete.

lokhor:
PR #1299 opened implementing this issue.

**Summary of deliverables:**
- `scripts/publish_launch_validation_evidence.py` — unified script: raw harness JSON → normalise → publish → dashboard
[...]

Issue #1287, opened 2026-06-17
Launch validation: full S21 evidence run and targeted S23U smoke


NickMonrad:
### S21 Tracked Reference (SM-G991B, Exynos)

**S21 full launch-scope evidence pass completed and merged.**

| Evidence | Status | Link |
|---|---|---|
| S21 full evidence report | ✅ Merged #1292 | `docs/test-triage/evidence/1287-s21-launch-validation-2026-06-18.md` |
| S21 model readiness | ✅ Documented | 38 skills pass, 1 known model degradation after ~45 sequential tests |
| Evidence inconsistencies reconciled | ✅ Merged | Suite sizing, slot_fill reclassification, xfail count, CI green |

**Key findings:**
- S21 (Gemma 4 E-2B): 38/38 skills pass at launch-slice scope
- 13-second model download latency (not a degradation)
- Known model degradation: after ~45 sequential tests, LLM routing collapses to single intent — documented as launch-deferred limitation
- Harness bug fixed: `device.py` missing imports

**Deferred to S23U reference:**
- S23U reference device (SM-S918B, Gemma 4 E-4B) connected via ADB wireless
- Targeted smoke/comparison testing not yet performed — blocked pending this issue owner
- Publishing workflow ready via PR #1299

### Publication workflow

PR #1299 (`feature/1295-launch-validation-evidence`) adds CLI tooling for publishing launch validation evidence:
- Normalises raw harness JSON to dashboard schema
- Publishes bundle to `test-results` branch
- Generates enriched Markdown report with phase breakdown
- Supports both CI and on-device evidence sources

**Do not close this issue** — remains open until S23U targeted evidence is captured and published.

lokhor:
## S21 launch validation phase complete

S21 full launch-scope validation has been run.

- **Device:** Samsung Galaxy S21 (SM-G991B) via USB
- **Commit/build tested:** `e51546df` — Debug variant (`assembleDebug`)
- **Commands:**
  ```bash
  ANDROID_SERIAL=R5CR605B71K ADB_WAIT_SECONDS=15 \
    python3 scripts/adb_skill_test.py \
    --exclude-tags destructive,device_state --model-readiness

  ANDROID_SERIAL=R5CR605B71K ADB_WAIT_SECONDS=20 \
    python3 scripts/adb_skill_test.py \
    --exclude-tags destructive,device_state --start-phase navigation --model-readiness
  ```
[...]

PR #1299, opened 2026-06-19
feat(#1295): add launch validation evidence publication workflow

lokhor:
## Summary

Implements #1295 — publish raw harness JSON and dashboard evidence for launch validation.

#### Commit `842f05d7` (first round)
The previous version only published the normalised evidence JSON via `--input`. This fix publishes the full evidence bundle via `--input-dir` and addresses all review findings:

- **Full evidence bundle**: raw JSON copied to `raw/` subdirectory alongside normalised JSON, CSV, and summaries
- **Bundle publishing**: changed from `--input <single_json>` to `--input-dir <out_dir>`
- **not_reached / excluded accuracy**: added CLI args with `not provided by source evidence` fallback
- **Phase breakdown**: derived from raw harness `results[].phase`, carried through normalised schema
- **Schema update**: optional `phase` string on case items
- **S23U documentation**: model corrected to Gemma 4 E-4B, daily driver warning strengthened
- **Docs updated**: bundle structure, directory publish mode, new CLI flags documented
- **Schema JSON fixed**: restored missing `"properties"` wrapper in cases items schema

#### Commit `edb92996` (second round — CHANGES_REQUESTED fixes)

1. **`--dry-run` guard**: `--dry-run` now exits immediately after printing the plan. It no longer runs the normaliser, copies raw JSON, generates Markdown, or calls `publish_test_evidence.py`. Added 4 unit tests in `test_publish_launch_validation_evidence.py` verifying: exit code 0, plan output, no evidence files created, and tolerance for unknown suite types.

2. **S23U Markdown fix**: The targeted-smoke bash block is now properly closed before the `**Publishing:**` heading. The publishing command lives in its own separate fenced bash block. The section renders correctly as Markdown.

3. **Raw JSON durability clarified**:
   - Added note in the *Published bundle* section: `raw/` subdirectory is local-only; to preserve raw JSON across runs, commit the evidence bundle directory or archive it externally
   - Added *Known limitations* entry: raw JSON is **not** pushed to `test-results` branch; must be committed/archived separately for durable storage
   - `test-results` branch stores only normalised evidence (`.json`, `.csv`, `.md`) for dashboard rendering

#### Commit `3737dda4` (this round — dry-run control flow regression fix)

1. **Dry-run control flow fix**: The `print("  SKIP: --dry-run flag set")` and `return` from the previous round were accidentally placed **outside** the `if args.dry_run:` block. This caused every invocation (including real publishing) to exit before reaching normalisation, raw JSON copy, Markdown generation, and the publisher. Fixed by indenting both statements inside the `if args.dry_run:` block.

2. **Non-dry-run regression test**: Added `NonDryRunTest.test_non_dry_run_reaches_normaliser` — imports the module with patched `sys.argv` and mocked `subprocess.run` to capture command construction. Verifies that: the normalisation subprocess command is invoked exactly once; the publisher subprocess command is also invoked exactly once; the full normalise → raw JSON copy → markdown → publish pipeline runs to completion for non-dry-run invocations.

### Generated/published artifact paths
| Artifact | Location | Published to dashboard? |
|---|---|---|
| Raw harness JSON | `raw/{report}.json` | **No** (local-only — commit separately for durable storage) |
| Normalised evidence | `{report}_evidence.json` | Yes |
| Case CSV | `{report}_cases.csv` | Yes |
| Normalised summary | `{report}_summary.md` | Yes |
| Launch validation summary | `launch-validation-summary.md` | Yes |

### #1287
**Remains open** until S23U targeted evidence is captured and published using this workflow.

### Validation
- [x] **196 Python unit tests (all pass)** — includes 5 publish-evidence tests (4 dry-run + 1 non-dry-run regression)
- [x] `--dry-run` produces plan output, exits 0, creates no files
- [x] Non-dry-run reaches normalisation, raw JSON copy, Markdown generation, and publisher
- [x] CI and Docs Drift Check currently green
- [x] No S23U device testing performed or required for this PR

**Do not merge without human review.**

### DIFF
--- a/docs/test-triage/evidence/1287-s21-launch-validation-2026-06-18.md
+++ b/docs/test-triage/evidence/1287-s21-launch-validation-2026-06-18.md
@@ -99,11 +99,15 @@ The S21 exhibits a consistent degradation pattern when running >45 tests sequent
   - Targeted daily-driver smoke: alarms, timers, weather, slot-fill, media (with Spotify installed), navigation, voice interactions
 - **orchestrator_recovery** tests — never reached (~12 tests due to timeout)
 - **false_positives** tests — never reached (~15 tests due to timeout)
-- **Dashboard / raw JSON evidence publication** — not performed. This PR publishes only the Markdown evidence summary.
-  - Raw harness JSON output is available from the ADB test runs for any follow-up tooling
-  - Dashboard publication would require either:
-    - Manual evidence entry into the test dashboard UI, or
-    - A follow-up automation script to ingest the harness JSON output
+  - **Dashboard / raw JSON evidence publication** — not performed. This PR publishes only the Markdown evidence summary.
+  - Raw harness JSON files from this run were on the ADB test device and have not been preserved.
+    They are **not recoverable** — the evidence pipeline was not yet in place to capture them durably.
+  - The launch validation evidence publication workflow has since been implemented
+    (issue [#1295](https://github.com/NickMonrad/kernel-ai-assistant/issues/1295)):
+    - `scripts/publish_launch_validation_evidence.py` chains harness output → normalise → publish
+    - See `docs/testing/launch-validation-evidence.md` for documentation
+    - All future launch validation runs should use this workflow to produce durable raw JSON,
+      normalised evidence, and dashboard entries
 
 ### #1287 Status
 
--- a/docs/testing/launch-validation-evidence.md
+++ b/docs/testing/launch-validation-evidence.md
@@ -1,0 +1,190 @@
+# Launch Validation Evidence
+
+How to run launch validation on physical devices and publish the results as
+durable evidence for the dashboard, PR summaries, and release snapshots.
+
+## Overview
+
+Launch validation verifies the on-device model and skill routing work correctly
+on a real Android device. Evidence is captured as:
+
+1. **Raw harness JSON** — full per-case results from `scripts/adb_skill_test.py`
+2. **Normalised evidence JSON** — transformed to the standard schema (see `test-evidence-schema.md`)
+3. **Markdown summary** — human-readable overview with metadata
+4. **Dashboard entry** — published to GitHub Pages via the `test-results` branch
+
+## Pipeline
+
+```
+ADB device → adb_skill_test.py → scripts/test-reports/{ts}_skills.json
+                                        ↓
+              scripts/publish_launch_validation_evidence.py
+                        ├── normalise_skills_report.py
+                        ├── publish_test_evidence.py
+                        └── docs/test-triage/evidence/{date}/{device}/
+```
+
+### Sequence
+
+1. **Run the harness** on the device.
+2. **Publish evidence** from the raw output.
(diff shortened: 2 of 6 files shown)


NickMonrad (changes requested):
Thanks — this is much closer and the main `--input-dir` publication path is now there. CI and Docs Drift Check are green on `842f05d7`.

I’m still requesting changes for two blockers:

1. `--dry-run` is not actually dry-run safe. In `publish_launch_validation_evidence.py`, the dry-run block prints the plan, but then execution continues into normalisation, raw-copying, Markdown generation, and publication. Please return/exit immediately after the dry-run output, before running the normaliser or publisher.

2. The S23U documentation has a broken Markdown code fence. The targeted-smoke bash block starts before the ADB command, but it is not closed before `**Publishing:**`, so the publishing heading and following fenced block render incorrectly. Please close the first code block after `--exclude-tags destructive,device_state --model-readiness` and keep the publishing example in its own fenced block.

One non-blocking point to clarify in docs/PR text: raw JSON is copied into the local evidence bundle under `raw/`, but `publish_test_evidence.py --input-dir` only publishes top-level `.json`, `.csv`, and `.md` files to `test-results`; it intentionally does not publish the `raw/` subdirectory. Please be explicit that raw JSON becomes durable only when the evidence bundle directory is committed/archived separately, or extend the publisher if the intended durability path is the `test-results` branch.

No S23U device testing is needed for this PR. This is tooling/docs only; S23U targeted smoke should wait until this publication flow is corrected and merged.

NickMonrad (changes requested):
Thanks — the docs fixes and raw JSON durability clarification look good, and CI/Docs Drift are green. I’m still requesting changes because the dry-run fix introduced a new blocker in the main execution path.

In `scripts/publish_launch_validation_evidence.py`, `print("  SKIP: --dry-run flag set")` and `return` are currently outside the `if args.dry_run:` block. That means **every non-dry-run invocation returns before normalisation, raw JSON copying, Markdown generation, and publishing**. The dry-run tests pass, but the real publish path is now disabled.

Please move the `SKIP` print and `return` inside the `if args.dry_run:` block, then add or update a test that exercises the non-dry-run path far enough to prove the normaliser/publisher command path is reached. A mocked/subprocess-based test is fine; no device testing is needed.

The S23U Markdown fence is now fixed and the raw JSON durability behaviour is documented clearly. No S23U testing is needed for this PR.

NickMonrad (approved):
Re-reviewed latest head `eebdc259`.

The previous blocker is fixed: `SKIP: --dry-run flag set` and `return` are now inside the `if args.dry_run:` block, and non-dry-run execution continues into normalisation, raw JSON copy, Markdown generation, and `publish_test_evidence.py --input-dir` publication.

I also checked the added regression coverage. The new non-dry-run test mocks subprocess execution and verifies the normaliser and publisher commands are each reached once, which covers the control-flow regression that blocked the last review.

CI and Docs Drift Check are green. The S23U docs fence and raw JSON durability clarification from the prior review remain addressed. No S23U testing is required for this tooling/docs PR.

Approved for merge.

PR #1304, opened 2026-06-20
feat(#1297): reject invalid slot values before dispatching tools

### LINKED ISSUE #1297: Reject invalid slot values before dispatching tools
Parent/related: #1014, #1287
Evidence source: PR #1292, `docs/test-triage/evidence/1287-s21-launch-validation-2026-06-18.md`

## Problem

The S21 launch validation evidence records a negative slot-fill safety test where an invalid timer slot value, `donuts`, dispatched to `set_timer` instead of being rejected or clarified.

This is currently classified as an xfail/known limitation in the evidence, but it is still a product quality and safety gap. Tool dispatch should not proceed when required slots contain invalid values.

## Scope

Improve slot validation before tool execution.

Start with timer/alarm-style slots and generalise where practical:

- duration values;
- times and dates;
- contact recipients;
- app names;
- locations / nearby queries;
- numeric fields where malformed natural-language values can be misread.

## Desired behaviour

When a slot value is invalid or unparseable:

- do not dispatch the tool;
- ask a clarification question where appropriate;
- return a safe validation error where clarification is not appropriate;
- record enough detail in logs/test output to explain why dispatch was blocked.

Examples:

- `set a timer for donuts` should not call `set_timer`.
- `send a message to purple` should not send or prepare a message without a resolvable contact.
- `navigate to yesterday` should not launch navigation without a valid destination.

## Acceptance criteria

- Add validation for timer duration slots so invalid values like `donuts` do not dispatch `set_timer`.
- Add or update negative slot-fill harness coverage for invalid duration values.
- Review other high-risk slot types and add validation where a low-effort common helper is possible.
- Ensure slot-fill tests report `validation_error`, `clarification_required`, or equivalent rather than `wrong_tool` for invalid slot values.
- Update the xfail in the harness if the negative test now passes.

## Launch posture

Launch-deferred unless manual smoke shows invalid slot dispatch can trigger destructive or surprising real-world actions. Prioritise before expanding more tool families.

lokhor:
## Summary

Add validation guards so invalid slot values are rejected before tool execution across **all** dispatch paths — both the ChatViewModel fast paths AND the LiteRT native SDK tool call path (`KernelAIToolSet.runIntent`).

## Architecture

Validation happens at two layers that share the same `SlotValidationRegistry`:

| Layer | Where | Guards |
|---|---|---|
| **ChatViewModel** (`validateBeforeDispatch`) | `ChatViewModel.kt` | QIR regex/classifier matches, recovery/anaphoric/intent resolution paths |
| **RunIntentSkill** (`execute()`) | `RunIntentSkill.kt` | **KernelAIToolSet.runIntent → executeSkill → run_intent** (the LiteRT native SDK tool path) |

Both call `SlotValidationRegistry.validateParams(intentName, params)` and fail closed before `NativeIntentHandler.handle()` on invalid values.

### Why two layers?

- **ChatViewModel** guards happen before tool dispatch — the user gets a clarification response, and the tool call is never made.
- **RunIntentSkill** guards at the skill boundary catch any caller that routes through `run_intent`, including:
  - `KernelAIToolSet.runIntent()` — the LiteRT native SDK `@Tool` method
  - Any future dispatch path that calls `skill.execute("run_intent", ...)`

### What's guarded

| Dispatch path | Guard |
|---|---|
| QIR regex/classifier direct match | ChatViewModel |
| Slot-fill completion | Already validated (SlotFillerManager) |
| Recovery confirmation | ChatViewModel |
| Confirmation fast-path | ChatViewModel |
| Recovery orchestrator Execute | ChatViewModel |
| Anaphoric intent ("do that") | ChatViewModel |
| **KernelAIToolSet.runIntent (LiteRT SDK tool)** | **RunIntentSkill (NEW)** |

### Built-in validators

| Validator | Intent.Slot | Rejects |
|---|---|---|
| `TimerDurationValidator` | `set_timer.duration_seconds` | "donuts", blank, "0", "-5", "86401" (over 24h) |
| `AlarmTimeValidator` | `set_alarm.time` | "donuts", "later", any unparseable time |
| `AlarmDayValidator` | `set_alarm.day` | Non-day strings (optional slot) |
| `WeatherLocationValidator` | `get_weather.location` | Blank/noise only |
| `NonEmptyStringValidator` | `set_timer.label` | Blank |

## Before/After

**Before:**
Model calls: runIntent(intentName="set_timer", parameters='{"duration_seconds":"0"}')
→ KernelAIToolSet.runIntent → executeSkill("run_intent", ...)
→ RunIntentSkill.execute → NativeIntentHandler.handle("set_timer", {duration_seconds: "0"})
→ Timer scheduled for 0ms (never fires)

**After:**
Model calls: runIntent(intentName="set_timer", parameters='{"duration_seconds":"0"}')
→ KernelAIToolSet.runIntent → executeSkill("run_intent", ...)
→ RunIntentSkill.execute
→ SlotValidationRegistry: set_timer.duration_seconds="0" invalid
→ SkillResult.Failure("run_intent/set_timer", "Sorry, I didn't understand that duration.")
→ KernelAIToolSet returns {error: ...} to the model
→ Model sees error, asks user for a valid duration

## Tests

./gradlew :core:skills:testDebugUnitTest (1620+ tests pass)
./gradlew :feature:chat:testDebugUnitTest (481 tests pass)
./gradlew assembleDebug (clean build)

### New: RunIntentSkillTest (15 tests)
Covers the shared boundary guard directly:
- set timer zero seconds is rejected — duration_seconds=0 → Failure, handler NOT called
- set timer negative seconds is rejected — duration_seconds=-5 → Failure, handler NOT called
- set timer over 24 hours is rejected — duration_seconds=90001 → Failure, handler NOT called
- set timer 5 minutes dispatches — duration_seconds=300 → calls handler with correct args
- set timer 30 seconds dispatches — duration_seconds=30 → calls handler
- set alarm invalid time is rejected — time="later" → Failure, handler NOT called
- set alarm valid time dispatches — time="5pm" → calls handler
- set alarm with valid time and day dispatches — time="7am" day="monday" → calls handler
- toggle flashlight passes through without validation — no registered validator → calls handler
- missing intent name returns failure — empty args → Failure, handler NOT called
- empty string intent name returns failure — intent_name="" → Failure, handler NOT called
- weather blank location is rejected — location="" → Failure, handler NOT called
- weather auckland location dispatches — location="Auckland" → calls handler
- first invalid param is returned before checking others — both invalid, first wins
- run_intent skill name prefix in failure result — skillName = "run_intent/set_timer"

### Existing test coverage (kept from earlier rounds)
- SlotValidationRegistryTest: 18 tests — individual validators, validate(), validateParams()
- SlotValidationRegistryDispatchTest: 14 tests — validateParams() edge cases, run_intent resolution
- SlotFillerManagerTest: 21 tests — slot-fill flow, retry-until-valid, blank-cancellation
- ChatViewModelInitTest/RecoveryTest/SettingsApplyTest/VoiceTest — updated with mock SlotValidationRegistry

## Notes

- The SlotValidationRegistry is lazily initialized in RunIntentSkill (by lazy) to avoid KSP/Hilt injection-order issues when co-resolving NativeIntentHandler in the same constructor.
- No S23U device testing was performed — it is the daily driver and explicitly out of scope.
- Closes #1297

### DIFF
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/RunIntentSkill.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/RunIntentSkill.kt
@@ -1,6 +1,7 @@
 package com.kernel.ai.core.skills
 
 import com.kernel.ai.core.skills.natives.NativeIntentHandler
+import com.kernel.ai.core.skills.slot.SlotValidationRegistry
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -21,6 +22,7 @@ class RunIntentSkill @Inject constructor(
     private val handler: NativeIntentHandler,
 ) : Skill {
 
+    private val validationRegistry: SlotValidationRegistry by lazy { SlotValidationRegistry() }
     override val name = "run_intent"
     override val description =
         "Perform a native Android device action. Supports flashlight, alarm, timer, calendar, email, SMS, " +
@@ -287,6 +289,14 @@ class RunIntentSkill @Inject constructor(
         val intentName = call.arguments["intent_name"]?.takeIf { it.isNotBlank() }
             ?: return SkillResult.Failure(name, "Missing required parameter: intent_name.")
         val extraParams = call.arguments - "intent_name"
+        // Validate slot values before dispatching — catches invalid already-populated
+        // params from direct model tool calls (KernelAIToolSet.runIntent) that bypass
+        // ChatViewModel.validateBeforeDispatch, and recovery/anaphoric paths
+        val invalid = validationRegistry.validateParams(intentName, extraParams)
+        if (invalid != null) {
+            val message = invalid.errorMessage ?: "Invalid value for $intentName. Please try again."
+            return SkillResult.Failure("$name/$intentName", message)
+        }
         return handler.handle(intentName, extraParams)
     }
 }
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/intent/IntentContractRegistry.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/intent/IntentContractRegistry.kt
@@ -158,6 +158,53 @@ class IntentContractRegistry {
                 riskLevel = IntentRiskLevel.MEDIUM,
             ),
 
+            // ── LOW RISK (alarms — time required, day optional) ───────────
+            IntentContract(
+                intentName = "set_alarm",
+                capability = "Set Alarm",
+                requiredSlots = mapOf(
+                    "time" to SlotSpec(
+                        name = "time",
+                        promptTemplate = "What time would you like the alarm set for?",
+                    ),
+                ),
+                optionalSlots = mapOf(
+                    "day" to SlotSpec(
+                        name = "day",
+                        promptTemplate = "Which day should the alarm be for?",
+                    ),
+                ),
+                riskLevel = IntentRiskLevel.LOW,
+            ),
+            IntentContract(
+                intentName = "set_timer",
+                capability = "Set Timer",
+                requiredSlots = mapOf(
+                    "duration_seconds" to SlotSpec(
+                        name = "duration_seconds",
+                        promptTemplate = "How long would you like the timer for?",
+                    ),
+                ),
+                optionalSlots = mapOf(
+                    "label" to SlotSpec(
+                        name = "label",
+                        promptTemplate = "What should I call this timer?",
+                    ),
+                ),
+                riskLevel = IntentRiskLevel.LOW,
+            ),
+            IntentContract(
+                intentName = "open_app",
+                capability = "Open App",
+                requiredSlots = mapOf(
+                    "app_name" to SlotSpec(
+                        name = "app_name",
+                        promptTemplate = "Which app would you like to open?",
+                    ),
+                ),
+                riskLevel = IntentRiskLevel.LOW,
+            ),
+
             // ── LOW RISK (safe to execute with slots filled) ──────────────────
             IntentContract(
                 intentName = "add_to_list",
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/slot/SlotFillResult.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/slot/SlotFillResult.kt
@@ -13,6 +13,17 @@ sealed class SlotFillResult {
         val request: PendingSlotRequest,
     ) : SlotFillResult()
 
+    /**
+     * The user provided a value for the current slot, but it failed validation —
+     * re-prompt with targeted clarification.
+     *
+     * The slot-fill remains pending; the user's next reply retries filling the
+     * same slot with a corrected value.
+     */
+    data class InvalidSlot(
+        val request: PendingSlotRequest,
+    ) : SlotFillResult()
+
     /** User sent a blank reply or explicitly cancelled. */
     data object Cancelled : SlotFillResult()
 }
--- a/core/skills/src/main/java/com/kernel/ai/core/skills/slot/SlotFillerManager.kt
+++ b/core/skills/src/main/java/com/kernel/ai/core/skills/slot/SlotFillerManager.kt
@@ -10,14 +10,17 @@ import javax.inject.Singleton
  * **Flow:**
  * 1. [QuickIntentRouter.route] returns [QuickIntentRouter.RouteResult.NeedsSlot] when a
  *    regex-matched intent is missing a required parameter.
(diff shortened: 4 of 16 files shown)


NickMonrad (changes requested):
Thanks — this is a strong implementation for the multi-turn slot-fill reply path, and CI/Docs Drift are green. The `donuts` retry flow is covered well: `SlotFillerManager` normalises, validates, returns `InvalidSlot`, keeps the pending request alive, and allows a valid retry to complete.

I’m requesting changes for one blocker: invalid slots are still not consistently rejected *before tool dispatch*.

The new validation is wired into `SlotFillerManager.onUserReply`, but the direct matched-intent path still bypasses it. For example, QIR timer regexes extract `duration_seconds` directly for inputs such as `set timer for 0 seconds` or a duration over the 24h policy, return `RegexMatch`, and `ChatViewModel` immediately builds a `SkillCall` / `run_intent` call from `matchedIntent.params`. `NativeIntentHandler.setTimer()` then receives the invalid value directly. That means the new validator blocks the prompted reply case (`set a timer` → `donuts`) but not all invalid slot values before dispatch, which is the issue acceptance target and PR title.

Please add a dispatch guard for already-populated params before execution. A few acceptable approaches:

1. Validate in `QuickIntentRouter.tryMatchPatterns()` after `paramExtractor`, before returning `RegexMatch`; if any extracted required/registered slot is invalid, return a slot-fill/clarification path instead of `RegexMatch`.
2. Validate in `ChatViewModel` immediately before creating/executing the `SkillCall`, and append the targeted clarification instead of dispatching.
3. Inject `SlotValidationRegistry` into `RunIntentSkill` / `NativeIntentHandler` and fail closed with a user-facing clarification result before the native handler performs the action.

Whichever approach you choose, please add regression coverage for direct populated invalid values, for example:
- `set timer for 0 seconds` does not dispatch/schedule and asks for a valid duration.
- `set timer for 25 hours` or `duration_seconds=86401` does not dispatch.
- existing valid direct timer routes still dispatch.

The current tests cover the multi-turn slot-fill path well, so keep those. The missing coverage is the direct route / dispatch path.

NickMonrad (changes requested):
Re-reviewed latest head `6409168`.

The previous blocker is mostly fixed for the ChatViewModel fast paths: the direct matched-intent path now calls `validateBeforeDispatch(...)` before `skill.execute(...)`, and the same guard is also applied to recovery execute, recovery confirmation, confirmation fast-path, and anaphoric intent dispatch. The new `SlotValidationRegistry.validateParams(...)` and direct-invalid registry tests cover the intended invalid timer/alarm/weather cases.

I’m still requesting changes because one dispatch path remains unguarded: LiteRT native tool execution.

`KernelAIToolSet.runIntent(...)` parses the model-supplied `parameters` into `args` and then calls `executeSkill("run_intent", args)` directly. `executeSkill(...)` then immediately calls `skill.execute(...)`. `RunIntentSkill.execute(...)` still just extracts `intent_name`, removes it from the args, and forwards straight to `NativeIntentHandler.handle(...)`. `NativeIntentHandler.setTimer(...)` still accepts any integer value and schedules it, so invalid model-supplied values such as `duration_seconds=0` or `duration_seconds=90001` can still bypass the new ChatViewModel guard when the model uses the native SDK `runIntent` tool path.

That means the implementation now protects QIR/ChatViewModel dispatch, but not all tool dispatch paths, and the issue/PR goal is specifically to reject invalid slot values before tools execute.

Please move the final guard to a shared core boundary that all `run_intent` callers cross, or add equivalent validation there. The cleanest option is likely `RunIntentSkill.execute(...)`, because both ChatViewModel fast paths and `KernelAIToolSet.runIntent(...)` ultimately route through `run_intent`. Another acceptable option is guarding inside `KernelAIToolSet.executeSkill(...)` for `run_intent`, but `RunIntentSkill` is the better fail-closed boundary.

Please also add regression coverage for the unguarded path, for example:
- `KernelAIToolSet.runIntent("set_timer", "{\"duration_seconds\":\"0\"}")` returns a validation/clarification error and does not call the underlying native handler.
- `RunIntentSkill.execute(...)` with `intent_name=set_timer,duration_seconds=90001` fails before `NativeIntentHandler.handle(...)`.
- Valid `run_intent` timer calls still execute.

Minor note: the current slot-fill completion path in `ChatViewModel` still executes after `SlotFillerManager` validation without calling `validateBeforeDispatch(...)`. That is probably okay because the pending slot has just been validated, but the PR body says the guard runs before every `skill.execute()`. Either route it through the shared guard for consistency or soften that claim.

NickMonrad (approved):

Issue #1384, opened 2026-07-08
fix: add_to_list bare query 'add to my list' not routed to slot-fill on S21


lokhor:
## Problem

The bare query "add to my list" fails to route to `add_to_list` intent on S21. The app returns NO_MATCH, preventing slot-fill from ever initiating.

Passing test for comparison:
- "add eggs to my shopping list" → `add_to_list` ✅ (includes both params inline, uses deterministic QIR path)

Failing tests:
- "add to my list" (single-slot) → NO_MATCH ❌
- "add to my list" (multi-slot) → NO_MATCH ❌

## Evidence

S21 cumulative run after #1383:

| Test | Case ID | Query | Expected | Actual |
|---|---:|---:|---:|---:|
| 8 | (default) | "add to my list" | add_to_list (item=eggs, list_name=shopping list) | NO_MATCH |
| 10 | slot_fill_add_to_list_multi | "add to my list" | add_to_list (item=eggs, list_name=shopping list) | NO_MATCH |
| 9 | (default) | "add eggs to my shopping list" | add_to_list (item=eggs, list_name=shopping) | **add_to_list ✅** |

Test 9 is tagged `deterministic` — it matches via QIR regex because both `item` and `list_name` are explicit in the query. Tests 8 and 10 require the app to:
1. Recognize the bare list intent from "add to my list"
2. Initiate slot-fill (NeedsSlot) to collect `item` and `list_name`
3. Dispatch after receiving both slots

The app fails at step 1 — returns NO_MATCH before any slot-fill begins.

## Root cause

Either:
- QIR regex for `add_to_list` requires explicit item name (no pattern for "add to my/the/our list" without item)
- MiniLM classifier doesn't have sufficient confidence for this specific phrasing

## Scope

- Audit QIR add_to_list regex patterns
- Check whether MiniLM add_to_list training samples include bare "add to my list" variants
- Ensure fix doesn't break existing positive cases (test 9, plus any fixture-based add_to_list tests)

## Related

- #1375 (parent issue, narrowed to this bucket)
- #1383 (harness capture reliability — this is the first follow-up to use the fixed harness)
- Report: `scripts/test-reports/2026-07-07T23-18-38Z_skills.json`
- Fixture umbrella: #1296 (contact fixtures for SMS/email — separate from list routing)

## Labels
bug, qir, device:S21

lokhor:
## Fixed in PR #1385

[...]

Issue #1377, opened 2026-07-04
fix: email fixture not configured on S21 — email queries routed to create_calendar_event


lokhor:
### Failures

- `send an email to John about the project update` → `create_calendar_event` (fixture_missing)
- `email Sarah the meeting notes` → `create_calendar_event` (fixture_missing)

### Root cause

The email fixture is not configured on this device. The ADB harness sets up the contact alias fixture for SMS/calling but does not configure email accounts. When the model receives an email request, it falls back to `create_calendar_event` because the email sending tool is unavailable.

### Classification

- **Type:** Test setup gap — not a product bug
- **Device:** S21
- **Priority:** P3 — affects 2 tests in misc phase

### Suggested fix

Add email fixture setup to the ADB harness preflight, or configure an email account on the test device. Alternatively, update the test expectations to accept `create_calendar_event` when email fixture is missing.

### Reproduce

```bash
python3 scripts/adb_skill_test.py --serial R5CR605B71K --phases misc --iso
```

### Labels
bug, automated-testing, device:S21

### Sweep report
`scripts/test-reports/automation-sweep/2026-07-04-s21-non-permission-sweep-rerun.md`

lokhor:
## Post-#1383 update

All 3 harness trust fixes are in (or ready):

| PR | Fix | Status |
|---|---|---|
| #1379 | Stale logcat boundary marker | ✅ Merged |
| #1382 | Oracle probe dedup | ✅ Merged |
| #1383 | `adb logcat -d` polling for fresh capture | 🔄 Open, CI green |

### Updated S21 evidence (slot_fill phase, cumulative)

| Test | Query | Expected | Actual | Status |
|---|---:|---:|---:|---:|
| 7 (single-slot) | "send an email" | send_email | NO_MATCH | ❌ |
| 12 (multi-slot) | "send an email" | send_email | NO_MATCH | ❌ |

Both variants fail at the initial query stage — the app returns NO_MATCH before any slot-fill begins. The `contacts:email_contact_seed` fixture label exists in test metadata but the actual email account/fixture is not configured on the device.

### Next action options

Three options for resolving:

1. **Configure an email account on S21** — Add a Google Account or test email account via ADB. Most reliable product-path validation.

2. **Add harness-level email fixture setup** — Create an ADB seed step that configures a test email account. More complex but automated.

[...]

Issue #1375, opened 2026-07-04
fix: remaining slot-fill product/fixture gaps after harness capture fixes


lokhor:
## Current status — post #1379 / #1382 / #1383

This issue has been narrowed again. The original `get_time`/`NO_MATCH` evidence was contaminated by harness capture problems, but those harness issues are now fixed and validated.

### What has been ruled out

- #1379 fixed stale logcat ring-buffer contamination in `capture_fresh_logcat()` using a boundary marker.
- #1382 fixed the ADB oracle probe collision where both warmup and oracle used identical `chat_input` text (`"what time is it"`), causing app-level dedup to suppress the second probe.
- #1383 replaced the competing continuous `adb logcat` subprocess in `capture_fresh_logcat()` with `adb logcat -d` polling, eliminating the S21 USB ADB fresh-capture starvation issue.

### Latest S21 baseline after #1383

| Phase | Previous | Current | Interpretation |
|---|---:|---:|---|
| `slot_fill` | 3/14 pass | 8/14 pass | Single-slot positive slot-fill flows now dispatch correctly; 6 remaining failures are product/fixture gaps. |
| `orchestrator_recovery` | 5/10 pass, 2 harness failures, 3 xfail | 7/10 pass, 0 real failures, 3 xfail | Harness failures eliminated; remaining xfails are documented `missing_extractor` gaps. |

S21 validation report from #1383:

```text
scripts/test-reports/2026-07-07T23-18-38Z_skills.json
```

## Current remaining failures

The remaining `slot_fill` failures are now classified as **field_mismatch / product-or-fixture gaps**, not harness failures.

Known remaining buckets from #1383:

1. `send_sms` multi-slot flows
   - Likely requires deterministic contact fixture handling and/or contact-name resolution behaviour.

2. `send_email` multi-slot flows
   - Likely overlaps with the existing email/contact fixture setup gap.

3. `add_to_list` multi-slot flows
   - Likely list-name normalization or list fixture/state issue.

4. Repeated identical query cases
   - App-level dedup may suppress repeated identical slot-fill prompts or completion turns.
   - Needs separate classification: expected UX behaviour vs harness should vary prompt text.

## Problem to fix now

Now that single-slot positive cases dispatch correctly, the next work is to classify and fix the remaining 6 `slot_fill` failures by bucket rather than treating #1375 as one broad dispatch failure.

Expected behaviour after a completed multi-slot flow:

[...]

Issue #1296, opened 2026-06-19
Add or repair launch validation fixtures for media, contacts, navigation, and false-positive coverage


NickMonrad:
Parent/related: #1113, #1287
Evidence source: PR #1292, `docs/test-triage/evidence/1287-s21-launch-validation-2026-06-18.md`

## Problem

The S21 launch validation evidence includes many failures that are fixture or harness limitations rather than clear product defects.

Examples include:

- media controls requiring an active playback context;
- Spotify/Netflix/Google Maps or equivalent apps not being available on the test device;
- contact-based SMS/call tests requiring seeded contact fixtures;
- navigation tests requiring location/GPS context;
- false-positive and orchestrator recovery phases not being reached before timeout.

Without reliable fixtures, launch validation produces noisy failures and makes it harder to distinguish product regressions from test environment gaps.

## Scope

Improve test fixture setup and classification for launch validation harness runs.

Areas to cover:

- Media fixture setup for play/pause/stop/skip/previous tests.
- Installed app assumptions for Spotify, Plex/Plexamp, Netflix, YouTube/YouTube Music, Google Maps, or documented substitutes.
- Contact fixture setup for SMS, email, and call flows.
- Location/GPS fixture strategy for navigation and nearby queries.
- False-positive phase execution as an independent phase.
- Clear skip/xfail behaviour when a fixture is intentionally unavailable.

## Desired behaviour

Harness evidence should clearly distinguish:

- product failure;
- fixture missing;
- app not installed;
- permission missing;
- device-state unavailable;
- accepted known limitation;
- xfail.

A missing fixture should not appear as a misleading wrong-tool product regression unless the test explicitly expects routing to work without the fixture.

## Acceptance criteria

- Add a documented launch fixture checklist for S21 and S23U.
- Add preflight checks that detect missing apps, contacts, permissions, and device state before running affected phases.
- Mark fixture-dependent tests as skipped/xfail with a clear reason when preconditions are missing.
- Ensure media-control tests have an active playback context or skip cleanly.
- Ensure contact-based tests use a seeded test contact or skip cleanly.
- Ensure navigation/location tests have known permissions and location preconditions or skip cleanly.
- Re-run a representative S21 harness slice showing cleaner classification.

## Notes

This issue is about evidence quality and test reliability, not necessarily fixing app routing logic. Product bugs exposed after fixtures are reliable should be raised separately.

NickMonrad:
[...]

PR #1385, opened 2026-07-09
fix(#1384): route bare 'add to my list' to add_to_list slot-fill

lokhor:
## Problem

Bare "add to my list" fails as NO_MATCH before slot-fill starts, while "add eggs to my shopping list" works (deterministic QIR path with inline params). Two root causes:

### 1. ActionsScreen LaunchedEffect race (product fix)

When `slot_reply_input` arrives in the same composition frame as `quick_action_input`, `executeAction()` dispatches its work via `viewModelScope.launch`, which hasn't set `_pendingSlot` by the time `LaunchedEffect(adbSlotReply)` fires. `onSlotReply` sees `_pendingSlot == null` and silently drops the reply.

### 2. Harness `keep_foreground=True` interference (harness fix)

`capture_fresh_logcat` with `keep_foreground=True` calls `_tap_keepalive()` (KEYCODE_WAKEUP) every 0.5s during the 5-second intermediate slot-reply poll. On S21 USB ADB, repeated wakeup events disrupt the slot bottom sheet, causing the pending slot state to be lost before the final reply is sent.

## Changes

| File | Change |
|---|---|
| `ActionsScreen.kt` | Add 50ms `delay()` in `LaunchedEffect(adbSlotReply)` to let `executeAction`'s coroutine prime `_pendingSlot` before processing the reply |
| `runners.py` | Set `keep_foreground=False` for intermediate slot replies in both old and new runners |
| `QuickIntentRouterSmokeTest.kt` | Add `"add to my list" → add_to_list` smoke test case |

## Validation (S21 cumulative)

| Phase | Before | After | Change |
|---|---:|---:|---|
| slot_fill | **8/14** pass | **12/14** pass | **+4 passes** |
| orchestrator_recovery | **7/10** pass, 0 fail, 3 xfail | **7/10** pass, 0 fail, 3 xfail | Unchanged |

### Slot_fill detail

| Test | Before | After | Reason |
|---|---:|---:|---:|
| 8 — "add to my list" (single) | ❌ | ✅ | Race + keep_foreground fixed |
| 10 — "add to my list" (multi) | ❌ | ✅ | Same |
| 11 — "send a message" (multi) | ❌ | ✅ | keep_foreground fixed |
| 12 — "send an email" (multi) | ❌ | ✅ | keep_foreground fixed |
| 6 — "send a message" (single) | ❌ | ❌ | Contact fixture `family_seed` not seeded (#1296) |
| 7 — "send an email" (single) | ❌ | ❌ | No email account on device (#1377) |

## Remaining failures after this PR

- 2 contact fixture gaps tracked by #1296 and #1377
- 3 orchestrator xfails (documented `missing_extractor` gaps)

## CI needed

Refs #1384, #1375

### DIFF
--- a/core/skills/src/test/java/com/kernel/ai/core/skills/QuickIntentRouterSmokeTest.kt
+++ b/core/skills/src/test/java/com/kernel/ai/core/skills/QuickIntentRouterSmokeTest.kt
@@ -92,6 +92,7 @@ class QuickIntentRouterSmokeTest {
             // add_to_list
             Arguments.of("add milk to my shopping list", "add_to_list"),
             Arguments.of("add eggs to my grocery list", "add_to_list"),
+            Arguments.of("add to my list", "add_to_list"),
         )
     }
 }
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsScreen.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ActionsScreen.kt
@@ -305,8 +305,16 @@ fun ActionsScreen(
 
     // ADB harness: deliver slot reply when slot_reply_input extra is provided.
     // onSlotReply guards internally — no-op if no slot is pending.
+    // Small delay (50ms) mitigates a race: when slot_reply_input arrives in the same
+    // composition frame as quick_action_input, executeAction's viewModelScope.launch
+    // coroutine (which primes _pendingSlot) hasn't run yet. Yielding the Main thread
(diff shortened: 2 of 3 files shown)


lokhor:
**Follow-up tracked at #1386** — the `delay(50L)` in `ActionsScreen.kt` is a pragmatic ADB-only race mitigation, not a deterministic sequencing model. #1386 proposes replacing it with a logcat-marker handshake or ordered test-only queue. Does not block this merge.

Issue #1402, opened 2026-07-14
Reproduce S21 post-idle Hey Jandal wake miss and missing acknowledgement cue


NickMonrad:
Parent umbrella: #65
Source performance investigation: #1391
Related silence/replay optimisation: #1398
Release tracker: #1014

## Problem

During the current-build paired enabled evidence run for #1391, Hey Jandal remained active for four hours on both devices with no crash, reboot or service loss. The post-idle functional check then produced different outcomes:

- S23 Ultra activated on the first spoken attempt.
- S21 did not visibly register the first attempt and activated on a retry.
- Earlier S21 evidence also recorded an apparently failed first attempt followed by a successful retry where no audible acknowledgement cue was heard.

The available evidence does not establish whether this is:

1. a wake-word detection/recall miss after extended silence;
2. a successful detector event whose cue was inaudible or routed incorrectly;
3. a wake-to-command handoff/session-start failure;
4. ordinary single-attempt variance.

This must be investigated separately from battery attribution and separately from speculative inference optimisation.

## Goal

Determine whether S21 has a reproducible post-idle wake/cue reliability problem on the current build, identify which stage fails, and either fix it or document evidence that it is not reproducible enough to block release.

## Investigation scope

Use the current merged debug build and preserve app data/models.

Run a focused physical-device test on S21, with a smaller S23 Ultra comparison where practical:

1. Confirm **Listen for Hey Jandal** is enabled and the service is active.
2. Confirm a normal awake/recently-armed wake interaction works before the idle period.
3. Leave the device locked and screen-off for a documented extended-idle period.
4. Attempt a natural “Hey Jandal” activation followed by a simple command.
5. Record separately whether:
   - `WakeWordDiag` recorded a high-confidence or verified activation;
   - the wake-word callback/session handoff occurred;
   - an acknowledgement cue was requested;
   - the cue was audible and used the expected audio route/volume;
   - STT capture began and accepted the command;
   - the service re-armed after completion.
6. Repeat enough times to distinguish a reproducible first-attempt failure from a one-off miss, without turning this into a battery benchmark.

Keep raw logs, device selectors, endpoints and unique identifiers private. Publish only sanitised evidence.

## Required distinctions

- A detector miss is not the same as a missing/inaudible cue.
- A detector activation with failed downstream handoff is not a wake-model recall failure.
- Do not change thresholds, provider selection, silence gating or model assets merely to make the symptom disappear.
[...]

PR #1431, opened 2026-08-01
fix(#1410): conform acoustic runner evidence to published schema

### LINKED ISSUE #1410: Execute the S21 post-idle acoustic wake reliability matrix
Parent design: #1403
Parent reliability investigation: #1402
Design source of truth: `docs/testing/wake-word-acoustic-reliability-harness.md` §15–§18 and §21 Slice F

## Goal

Use the completed acoustic source, target event journal, cue policy, evidence/dashboard support and paired runner to measure current-main wake reliability on the S21 and S23 Ultra and make the v1.0 release decision from observed behaviour.

This issue owns the final targeted regression/release gate. It must not broaden back into wake-model research unless the measured release result demonstrates a clearly unacceptable product failure rate.

## Current release baseline

The classifier-only remediation investigation in #1444 is complete and closed `not planned` for v1.0 after C2–C5, W1 and T1 failed to improve the committed classifier without unacceptable recall/false-positive/fixture regressions.

For the release matrix:

- use current `main` and the committed stable classifier;
- do not wait for another classifier candidate;
- do not change Stage 1, Stage 2, classifier thresholds, providers, verifier routing, source volume, placement or route to make the matrix pass;
- #1433 first-attempt STT remediation is complete;
- #1444 model experimentation is complete for v1.0.

## Preconditions

Before execution:

- #1405 cue audit/policy is complete and reviewed;
- #1406 controlled source is complete and physically feasible;
- #1407 target journal/event wait is complete;
- #1408 evidence schema/dashboard support is complete;
- #1409 paired runner is complete and reviewed;
- one fresh private natural “Hey Jandal” fixture and approved Qwen command fixture are available with hashes;
- S21 and S23U are running one exact current-main APK;
- all release-isolation and privacy tests pass;
- frozen source volume, placement and audio route are restored and recorded.

## Historical diagnostic baseline

The original `diagnostic_pre_fix` matrix remains historical evidence and must not be rewritten or reclassified. It demonstrated the defects subsequently remediated under #1432/#1433 and must remain available for comparison.

## Final release matrix

Run one complete independent `regression_post_fix` matrix against the current stable implementation.

Required valid S21 trials:

| Idle | Wake-only | Wake + command |
|---|---:|---:|
| 10 seconds / recently armed | 5 | 3 |
| 30 seconds | 5 | 0 |
| 2 minutes | 5 | 3 |
| 15 minutes | 2 | 0 |
| 30 minutes | 2 | 2 |

Total required valid S21 positions: **27**.

Required valid S23U comparison:

| Idle | Wake-only | Wake + command |
|---|---:|---:|
| 2 minutes | 15 | 2 |
| 30 minutes | 2 | 1 |

Total required valid S23U positions: **20**.

The original 8 S23U comparison positions remain part of this matrix. The additional 12 positions are 2-minute wake-only trials so the 95% reliability target can be measured directly without unnecessarily expanding long-idle or command-path coverage.

Invalid attempts remain visible and may be repeated only to complete required valid counts. Valid failures must never be retried away or replaced.

Classify every valid failure as:

- acoustic/gate miss;
- classifier/model miss;
- activation handoff failure;
- STT readiness failure;
- cue audio failure;
- cue audibility unconfirmed;
- command capture/routing failure;
- service re-arm failure;
- spontaneous activation / false positive;
- unclassified.

## v1.0 acceptance decision

Wake false negatives are accepted at different rates by device class. False positives remain a much tighter release constraint.

### S21 wake activation

Across the 27 required valid S21 positions:

- at least **22/27** must produce a valid intended wake activation (**≥81.5%**);
- therefore at most **five** genuine acoustic/classifier wake misses are acceptable;
- a wake activation is counted separately from downstream command execution so a command-path failure cannot be hidden as a wake-model failure.

This implements the v1.0 product decision that approximately 80% S21 wake reliability is acceptable. Do not alter the threshold after seeing the run result.

### S23 Ultra wake activation

Across the 20 required valid S23U positions:

- at least **19/20** must produce a valid intended wake activation (**≥95%**);
- therefore at most **one** genuine acoustic/classifier wake miss is acceptable;
- report the S23U numerator/denominator independently from the S21 result.

### False activations and system correctness

Acceptance still requires:

- **zero spontaneous/false wake activations on either device during controlled observation periods**;
- **zero unclassified valid failures**;
- **zero harness/system failures counted as valid product successes**;
- every successful wake reaches the expected callback/session path;
- detector re-arm is verified after completion/cancellation;
- cue behaviour satisfies the approved app-level and monitored acoustic standard.

False-positive tolerance is intentionally stricter than false-negative tolerance. Do not trade additional spontaneous activations for higher wake recall.

### Command trials

For every required wake-plus-command trial where the intended wake activates:

- the first normal STT path must reach readiness according to the merged #1433 behaviour;
- “What time is it?” must be captured and routed correctly in the correlated session/generation;
- retry-induced or runner-delay masking is not accepted as the normal successful path.

A permitted wake miss may occur in either a wake-only or wake-plus-command position; when the wake itself does not activate, command routing is recorded as not reached rather than as a second independent failure.

## Release decision

After the single complete matrix:

- **S21 ≥22/27, S23U ≥19/20, zero spontaneous activations, zero unclassified outcomes, and command-path acceptance satisfied:** wake reliability is accepted for v1.0; update and close #1402 and #1410.
- **S21 ≤21/27, S23U ≤18/20, any spontaneous activation, any unclassified valid failure, or command-path acceptance failure:** release gate fails. Do not automatically restart classifier research. Stop for an explicit product/architecture decision using the measured failure rate and classifications.

Do not run piecemeal replacement trials to improve either percentage.

## Evidence publication

Publish one normalised evidence record per target device through the existing `test-results` process, including:

- `regression_post_fix` run kind;
- matrix version;
- source/target context and exact commit/APK hash;
- fixture hashes;
- audibility-preflight record;
- valid pass/fail/invalid counts;
- intended wake activation numerator/denominator and percentage;
- spontaneous activation count;
- failure classifications;
- command-path success counts;
- timing summaries;
- sanitised drill-down artifacts.

Keep all raw audio, selectors, endpoints and private logs in gitignored local storage.

## Completion actions

- update #1402 with the final measured S21/S23U reliability evidence and release decision;
- close #1402 only after #1410 reaches an accepted release decision;
- update #1398/#1399 only when final evidence materially changes their priority or acceptance needs;
- retain the frozen wake matrix as reusable targeted regression evidence;
- do not reopen #1444 merely because an allowed S21 or S23U acoustic miss occurs.

## Acceptance criteria

- [ ] All preconditions and release/privacy gates are met.
- [ ] One exact current-main APK is used for the complete matrix on both devices.
- [ ] All 27 required valid S21 positions are completed without replacing valid failures.
- [ ] All 20 required valid S23U positions are completed without replacing valid failures.
- [ ] S21 intended wake activation is at least 22/27 (≥81.5%).
- [ ] S23U intended wake activation is at least 19/20 (≥95%).
- [ ] Zero spontaneous/false wake activations occur on either device during controlled observation periods.
- [ ] Zero valid failures remain unclassified.
- [ ] Every successful wake follows the expected activation/session path and detector re-arms correctly.
- [ ] Every wake-plus-command trial whose intended wake activates captures and routes “What time is it?” in the correlated normal STT path.
- [ ] Cue audibility and STT readiness meet the approved standards.
- [ ] Placement, source volume, fixture and route remain fixed.
- [ ] Evidence completeness and privacy validation pass.
- [ ] Normalised evidence is published and displayed correctly in the dashboard.
- [ ] #1402 is updated and resolved only after the final release decision.
- [ ] A failed release gate stops for explicit product/architecture review rather than automatically reopening classifier experimentation.

## Dependencies

Implementation dependencies #1405–#1409 and #1433 are complete. Classifier investigation #1444 is complete and closed `not planned` for v1.0. This issue is ready for final matrix execution once the current-main APK and frozen physical preflight are prepared.

lokhor:
## Summary

Harness-only fixes surfaced by the physical #1410 acoustic matrix execution (no production code changes).

1. **`public_environment_snapshot`** now normalises device-native values to the published evidence schema (`scripts/testdata/test_evidence.schema.json`): string `ringer_mode`/`dnd_mode` names and integer `uptime_seconds`. Previously the runner emitted int modes and float uptime, which failed schema validation on real devices (`schema:cases.0 … ringer_mode 0 is not of type 'string'`), blocking all evidence export for physical runs.
2. **Missing `import sys`** — the `__main__` fatal handler referenced `sys.stderr`/`SystemExit(2)` without importing `sys`, masking the schema error with a `NameError`.
3. Regression test `test_public_environment_snapshot_normalises_to_evidence_schema_types` locks the projection contract.

## Out-of-window command handling (reviews 4833215906 + 4833514166)

**Forward prevention (accepted, unchanged):** immediately before command dispatch the runner takes one target-journal snapshot; if it already contains the correlated session's terminal event (`SESSION_COMPLETED`/`SESSION_CANCELLED`), command playback is suppressed, the attempt is `INVALID` / `command_outside_listening_window`, and the slot stays eligible for bounded retry. No idle polling, no fixed delay. The pre-dispatch target observation is persisted per attempt (`target_timing.pre_dispatch_terminal_sequence`) as same-target ordering evidence.

**Retrospective reclassification now requires proof (remediation of review 4833514166):** the prior raw cross-device wall-clock comparison (`source request_wall_clock_ms > target terminal w`) is **removed** — two Android devices' epoch wall clocks are not a synchronised clock. Reclassification on checkpoint load / evidence export happens only when:

1. **Same-target pre-dispatch proof** — a persisted target-journal observation taken before command invocation already contained the correlated terminal sequence (authoritative, no cross-device clocks); or
2. **Persisted bounded clock-alignment evidence** — a per-run `clock_alignment` record (validated structurally and against the run's device pair: `offset_range_ms` [lo, hi] bounding target−source wall-clock offset, non-negative `uncertainty_ms`, non-empty `method`) makes the command provably after the terminal for every offset in range (`request + lo > terminal`).

Without either, the recorded classification is **preserved** — an unmeasured source-ahead offset must not silently remove a genuine product failure. Evidence priority 1→2→3 is implemented in `command_delivery_after_session_end()` / `reclassify_out_of_window_command_attempt()` with the runner-level `clock_alignment` checkpoint field.

## Physical evidence (regenerated from the preserved private run, no new trials)

The preserved S21 run contains **no pre-command target snapshot and no clock-alignment record**, so trial 007 (10s wake+cmd) keeps its recorded `FAILED / command_capture_or_routing_failure` classification:

- **S21 diagnostic: 27/27 valid — 15 passed / 12 failed / 8 invalid** (complete matrix); classifications 11× `classifier_model_miss` + 1× `command_capture_or_routing_failure`.
- **S23U comparison: unchanged 8/8 — 2 passed / 6 failed / 5 invalid**, all failures `classifier_model_miss`.

Raw fixtures, checkpoints and the preserved snapshot were not altered; totals derive from attempt records via the corrected runner path.

## Tests

`OutOfWindowCommandTests` now also covers: raw wall-clock ordering without calibration or marker is **not** reclassified; source-ahead/source-behind skew scenarios (the review's 750 ms-ahead example) are inconclusive even with a persisted range unless conclusive at every offset; same-target pre-dispatch markers and conclusive validated alignment records do reclassify; untrustworthy alignment records (bad range/uncertainty/method/wrong device pair) are rejected; uncalibrated preserved runs keep their classification on load; exported totals reconcile in both directions. Forward-path tests unchanged.

## Verification (head `195ae7d2`)

- `python3 -m unittest scripts.tests.test_acoustic_wake_reliability_runner`: **69 passed**; full `scripts/tests` discovery: **694 passed**.
- Runner fixture/self-test: PASS.
- Regenerated evidence: schema errors 0; `validate_record` privacy/schema valid; dashboard renders **27/27** (S21) and **8/8** (S23U).
- `git diff --check`: clean.

## Evidence

Physical run evidence remains private under `scripts/private-acoustic-runs/` (gitignored). Final totals and the trial-007 classification note are recorded on the updated #1410 and #1402 execution comments.

Closes #1410 evidence-export blocker only; the wake reliability remediation decision is recorded separately on #1402.

### DIFF
--- a/docs/testing/wake-word-acoustic-reliability-harness.md
+++ b/docs/testing/wake-word-acoustic-reliability-harness.md
@@ -540,6 +540,7 @@ Evaluate validity first.
 |---|---|
 | Source playback, route, hash or restoration failure | invalid: `source_stimulus_failure` |
 | Target screen/state/service/reboot/ADB/evidence boundary invalid | invalid: `device_environment_error` |
+| Command dispatched after the correlated session's terminal event | invalid: `command_outside_listening_window` |
 | Source succeeded but target saw no credible audio/voiced transition | `acoustic_or_gate_miss` |
 | Gate opened and inference became ready but no activation | `classifier_model_miss` |
 | Activation occurred but callback/session did not start | `activation_handoff_failure` |
--- a/scripts/acoustic_wake_reliability_runner.py
+++ b/scripts/acoustic_wake_reliability_runner.py
@@ -1320,24 +1320,86 @@ def classify_attempt(
 TERMINAL_SESSION_EVENT_TYPES = {"SESSION_COMPLETED", "SESSION_CANCELLED"}
 
 
-def command_delivery_after_session_end(attempt: MatrixAttempt) -> bool:
-    """Return whether recorded evidence proves command delivery began after
+def validated_clock_alignment(
+    value: Any,
+    *,
+    source_alias: str | None = None,
+    target_alias: str | None = None,
+) -> dict[str, Any] | None:
+    """Structurally validate a persisted source↔target wall-clock alignment record.
+
+    A record is trusted only when it is persisted with the private run, names
+    the paired devices, and bounds the offset: ``offset_range_ms`` [lo, hi] =
+    possible values of (target wall clock − source wall clock), a non-negative
+    ``uncertainty_ms``, and a non-empty ``method`` describing the measurement.
+
+    Two Android devices' epoch wall clocks are NOT treated as synchronised
+    without such a record; the range is the only basis for cross-device
+    ordering.  Returns the validated record or None when untrustworthy.
+    """
+    if not isinstance(value, dict):
+        return None
+    offset_range = value.get("offset_range_ms")
+    if not isinstance(offset_range, list) or len(offset_range) != 2:
+        return None
+    lo, hi = offset_range
+    if (
+        isinstance(lo, bool) or isinstance(hi, bool)
+        or not isinstance(lo, int) or not isinstance(hi, int)
+    ):
+        return None
+    if lo > hi:
+        return None
+    uncertainty = value.get("uncertainty_ms")
+    if (
+        isinstance(uncertainty, bool)
+        or not isinstance(uncertainty, int)
+        or uncertainty < 0
+    ):
+        return None
+    method = value.get("method")
+    if not isinstance(method, str) or not method.strip():
+        return None
+    if source_alias is not None and value.get("source_alias") != source_alias:
+        return None
+    if target_alias is not None and value.get("target_alias") != target_alias:
+        return None
+    validated = {
+        "offset_range_ms": [lo, hi],
+        "uncertainty_ms": uncertainty,
+        "method": method,
+    }
+    if "source_alias" in value:
+        validated["source_alias"] = value["source_alias"]
+    if "target_alias" in value:
+        validated["target_alias"] = value["target_alias"]
+    return validated
+
+
+def command_delivery_after_session_end(
+    attempt: MatrixAttempt,
+    clock_alignment: dict[str, Any] | None = None,
+) -> bool:
+    """Return whether recorded evidence PROVES command delivery began after
     the correlated target session ended.
 
-    Authoritative evidence, in order of strength:
-    - target-journal order: the correlated session's terminal event
-      (``SESSION_COMPLETED``/``SESSION_CANCELLED``) is present in the recorded
-      event path; and
-    - the command request wall clock (``request_wall_clock_ms``, source
-      ``System.currentTimeMillis()`` epoch ms) is later than the terminal
-      event's wall clock (target ``System.currentTimeMillis()`` epoch ms).
+    Evidence priority, no cross-device clocks compared directly:
 
-    Both timestamps are epoch wall clocks (a global, contract-defined clock
-    domain); device-relative ``elapsedRealtime`` values are never compared
-    across devices.  Wall-clock skew between the paired devices was verified
-    < 200 ms during the preserved runs via speech-detected/playback
-    alignment, so a recorded request that follows the terminal event by
-    hundreds of milliseconds is authoritative.
+    1. Same-target pre-dispatch proof: the runner persisted a target-journal
+       observation taken immediately before command invocation
+       (``target_timing.pre_dispatch_terminal_sequence``).  When that
+       observation already contained the correlated session's terminal
+       sequence, ordering is proven by the target journal alone.
+    2. Persisted bounded clock-alignment evidence: a per-run
+       ``clock_alignment`` record (validated by ``validated_clock_alignment``)
+       bounds (target − source) wall-clock offset.  The command request is
+       provably after the terminal event only when it is after it for every
+       offset in the recorded range:
+       request_wall_clock_ms + offset_range_lo > terminal_wall_clock_ms.
+    3. Unproven: return False and preserve the recorded classification.
+
+    Epoch wall clocks are a common unit, not a synchronised shared clock;
+    a raw cross-device comparison is never treated as proof.
     """
     if attempt.matrix_slot.wake_only:
         return False
@@ -1355,33 +1417,52 @@ def command_delivery_after_session_end(attempt: MatrixAttempt) -> bool:
                 terminal_wall_clock_ms = candidate
     if not isinstance(terminal_wall_clock_ms, int):
         return False
-    return request_wall_clock_ms > terminal_wall_clock_ms
+
+    # Priority 1: the target journal itself showed the terminal event in a
+    # snapshot taken immediately before command invocation.
+    pre_dispatch = attempt.target_timing.get("pre_dispatch_terminal_sequence")
+    if isinstance(pre_dispatch, int):
+        return True
+
+    # Priority 2: persisted bounded clock-alignment, conclusive at every
+    # offset in the recorded range.
+    alignment = validated_clock_alignment(clock_alignment)
+    if alignment is not None:
+        offset_lo = alignment["offset_range_ms"][0]
+        if request_wall_clock_ms + offset_lo > terminal_wall_clock_ms:
+            return True
+    return False
 
 
-def reclassify_out_of_window_command_attempt(attempt: MatrixAttempt) -> bool:
-    """Defensively invalidate a recorded command attempt whose delivery began
-    after the correlated target session ended.
+def reclassify_out_of_window_command_attempt(
+    attempt: MatrixAttempt,
+    clock_alignment: dict[str, Any] | None = None,
+) -> bool:
+    """Defensively invalidate a recorded command attempt whose delivery is
+    PROVEN to have begun after the correlated target session ended.
 
-    The harness had no opportunity to recognise the command because the
-    stimulus was delivered outside the valid listening window; such an
-    attempt must never be exported as a valid product failure.  Returns
-    whether the attempt was reclassified.
+    Reclassification happens only when authoritative evidence (same-target
+    pre-dispatch journal proof, or a persisted validated clock-alignment
+    record conclusive beyond its range) establishes the ordering.  Without
+    proof the recorded classification is preserved: uncalibrated cross-device
+    wall clocks are never treated as synchronised.  Returns whether the
+    attempt was reclassified.
     """
-    if not command_delivery_after_session_end(attempt):
+    if not command_delivery_after_session_end(attempt, clock_alignment):
         return False
     attempt.status = AttemptStatus.INVALID
     attempt.classification = None
     attempt.invalid_reason = InvalidReason.COMMAND_OUTSIDE_LISTENING_WINDOW
     detail = (
-        "command delivery began after the correlated target session ended "
-        "(session terminal event precedes the command request wall clock)"
+        "command delivery is proven to have begun after the correlated "
+        "target session ended (same-target ordering or validated clock alignment)"
     )
     if detail not in attempt.failures:
         attempt.failures.append(detail)
     attempt.invalid_details["listening_window_closed"] = {
         "evidence": (
-            "target journal terminal event precedes command request "
-            "wall clock in the epoch clock domain"
+            "proven by target-journal pre-dispatch observation or a persisted "
+            "validated source-target clock-alignment record"
         ),
     }
     return True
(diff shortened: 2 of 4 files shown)


NickMonrad (changes requested):
## Review result: changes required

I found **one material defect**.

### High — an out-of-window harness command is still exported as a valid product failure

**Location:** `scripts/acoustic_wake_reliability_runner.py`, `run_trial()`, after `classify_attempt()` and before the attempt is added to `valid_failed_slots`.

**Failure scenario:** the recorded S21 10-second `wake_plus_command` attempt invoked the command after the correlated `SESSION_COMPLETED` event. The operator evidence correctly says this was not a product command-recognition failure, but the runner still records it as `FAILED / command_capture_or_routing_failure`, treats the required matrix position as complete, and includes it in the valid-failure count and product pass rate.

**Why this is incorrect:** the acoustic evidence contract defines `failed` as a valid product failure. Harness orchestration that delivers the stimulus after the listening window has closed is an invalid attempt and must not reduce product reliability. The current result is internally contradictory: the published narrative says the case is not a product failure while the exported evidence and 15-pass/12-fail total say it is.

**Smallest remediation:**

- classify a command attempt as `invalid` when the correlated target session has ended before the harness delivers the command;
- give it an explicit harness/source-timing invalid reason and do not mark that matrix slot complete;
- add a focused regression test proving the invalid attempt remains visible and the slot remains eligible for its bounded retry;
- regenerate the sanitised evidence and correct the #1410/#1402 totals. Because this attempt currently supplies one of the claimed 27 valid S21 positions, either complete a compatible retry for that slot or report the matrix as 26/27 valid rather than retaining the 27/27 claim.

The schema type normalisation and missing `sys` import are otherwise correct, and CI plus Docs Drift Check pass at this head.

NickMonrad (changes requested):
## Re-review result: changes required

The original out-of-window classification finding is mostly addressed, but the remediation introduces **one new material defect**.

### High — raw cross-device wall clocks can reclassify genuine product failures as invalid

**Location:** `scripts/acoustic_wake_reliability_runner.py`, `command_delivery_after_session_end()` and its unconditional use from `load_checkpoint()` / `export_evidence()`.

**Failure scenario:** the source phone's wall clock is ahead of the target phone's clock. A command can be dispatched while the target session is still open, yet `source.request_wall_clock_ms > target.SESSION_COMPLETED.w` solely because of clock offset. On checkpoint load or evidence export, the runner then changes a genuine `FAILED / command_capture_or_routing_failure` into `INVALID / command_outside_listening_window`, removes it from the product failure count and leaves its matrix position incomplete.

For example, with the source clock 750 ms ahead, a command actually requested 300 ms before the target session ends appears 450 ms after it under the current comparison. `System.currentTimeMillis()` uses the same unit and epoch on both devices, but it is not a synchronised shared clock with a guaranteed offset. The code comments mention a <200 ms estimate for the preserved run, but no per-run offset/uncertainty is persisted, validated or enforced; the generic function applies a zero-margin comparison to every future run.

**Why this violates the evidence contract:** invalid attempts must represent proven harness/environment failures. This path can silently discard a real product failure and make a launch-gate result look better. The deterministic tests use perfectly aligned synthetic clocks and therefore do not cover the credible failure path.

**Smallest remediation:** keep the target-journal pre-dispatch check, which is authoritative and uses one target clock/order. For defensive post-run reclassification, require persisted per-run clock-offset/uncertainty evidence (or another single-clock/ordered marker) and only invalidate when the adjusted ordering is conclusive beyond that uncertainty. With no trustworthy calibration, leave the recorded classification unchanged rather than guessing. Add focused tests for source-ahead/source-behind skew and an uncalibrated preserved run. Regenerate the corrected physical evidence only through a path that represents and validates the preserved run's claimed clock bound.

The corrected 26/27 totals, scheduler retry behaviour, schema update and GitHub comment corrections are otherwise consistent, and CI plus Docs Drift Check pass at head `9cfa2609e6796e83fd2fb9714ae91da737f174bb`.

NickMonrad (approved):
## Focused re-review: approved

The clock-skew finding from review `4833514166` is resolved.

- Raw source/target wall-clock ordering is no longer treated as proof.
- Retrospective invalidation now requires either persisted same-target pre-dispatch journal ordering or a validated bounded per-run clock-alignment record.
- Uncalibrated and inconclusive cases preserve the original product-failure classification.
- The future pre-dispatch target-journal guard remains authoritative and avoids cross-device clocks.
- Source-ahead, source-behind, wrong-device-pair, uncalibrated checkpoint, scheduler and evidence-total paths are covered by focused tests.
- The preserved physical run is conservatively restored to 27/27 valid, 15 passed, 12 failed and 8 invalid; #1410 and #1402 agree with that result.
- CI and Docs Drift Check pass at this exact head.

No material defects remain in the remediation changes. PR #1431 is ready to mark ready for review and merge.

NickMonrad (commented):
## Post-merge focused review: follow-up required

I found **one material defect in the remediation changes**.

### High — the late-command check compares source request time, not playback start

**Location:** `scripts/acoustic_wake_reliability_runner.py`, `command_delivery_after_session_end()`, specifically the use of `attempt.source_timing["command"]["request_wall_clock_ms"]`; and `run_trial()`, where command timing persists `playback_start_monotonic_ms` but not the source helper's `playback_started.wall_clock_ms`.

**Concrete failure scenario:** the pre-dispatch target snapshot observes the session still open, then the target emits `SESSION_COMPLETED` while the source helper is preparing audio. The source request therefore precedes the terminal event, but physical playback begins afterwards. The source result contract already distinguishes `request_submitted`, `prepare_started`, `prepared`, and `playback_started`, each with wall-clock evidence. Because retrospective reclassification compares the earlier request timestamp, this attempt remains `FAILED / command_capture_or_routing_failure` even when calibrated evidence proves the audible command started after listening had ended.

**Why this violates the evidence contract:** `command_outside_listening_window` is defined by when the command is delivered, and the function itself claims to prove when delivery began. Using request submission leaves a credible race between the target snapshot and actual speaker onset, so harness timing can still be counted as a valid product failure.

**Smallest remediation:** preserve the existing target-journal pre-dispatch guard, persist the command source's `playback_started.wall_clock_ms` into `attempt.source_timing["command"]`, and use that timestamp—rather than `request_wall_clock_ms`—for the validated clock-alignment comparison. Add focused tests where request is before the target terminal but playback start is conclusively after it, plus the control case where playback start remains inside the window.

The prior cross-device clock-skew finding is otherwise resolved, and CI passed at the merged head.

PR #1441, opened 2026-08-03
test(#1432): pin classifier no-content floor and add episode capture-level diagnostics

lokhor:
Refs #1432

# Outcome 3 — evidence inconclusive.

The exact-head bounded smoke validated the capture-level diagnostic on both
devices but produced no no-candidate trial. The acoustic-capture/SNR hypothesis
is therefore neither confirmed nor refuted. This PR delivers diagnostics and
real-model boundary tests only; it does not deliver a production detector fix
or resolve #1432.

## Provenance

- **Starting main:** `b9360af5718f7dd4fd94c9177d3572e736e1c5a8` (origin/main, #1440 merge)
- **Final/current head:** `c516257569f27af8b961a6edf0104246bf45efbb` (branch `fix/1432-classifier-no-candidate`) — **unchanged before and after physical testing**
- **Exact smoke APK SHA-256:** `5142e2e172f7802ccb63821d60dcb7c3ecabcd1fc99f8fcbcb555b22034b1fdd` (built from the exact head)
- **App:** 0.1.0 / versionCode 1
- **Wake classifier:** `11bcdb0d…38c206` (pinned, in-APK verified)
- **Whisper verifier files** (both devices, verified): `0ce578b8…` encoder / `06c0e6ff…` decoder / `306cd27f…` tokens
- No APK, raw evidence or private audio was committed.

## Historical evidence that motivated the capture-level hypothesis

The preserved #1440 smoke contained three adjacent S23U-target trials:

- a no-candidate trial whose phrase episode max confidence was **0.001** (the real classifier's no-content floor) — no `ACTIVATION_CANDIDATE`;
- an adjacent **0.6335** LOW candidate that reached `VERIFIED_ACTIVATION` through the #1440 Whisper verifier;
- an adjacent **0.8993** HIGH activation.

An earlier #1438-era smoke preserved a fully pinned **0.3444** no-candidate trial
(`wakeworddiag-s23u.log`: `stage3Evals=2442 maxConfidence=0.34444994 maxConfidenceChunk=1649`).
Chunk 1649 is the phrase at its tail-aligned position (onset +0.9–1.0 s — the
position every passing trial fires at) with a complete chronological ~16-slot
window; the 2442 evaluations are continuous, so no frame was dropped or skipped.

Across the available historical journals, **no detector lifecycle divergence
was found** between the failing and passing trials: gate phase at phrase
arrival, VOICED/STAGE2_RESUMED/STAGE3_READY events, embedding-ring
[...]

PR #1448, opened 2026-08-08
test(#1410): retain wake acoustic evidence for reproducibility

lokhor:
Closes the missing-evidence prerequisite from the #1410 discrepancy investigation (decision **C — stimulus equivalence cannot be established**): the next bounded S23U→S21 reproducibility test must prove whether the S21 received materially comparable acoustic input to the `capture-1432` run.

**Remediated for review at `af8e1ace`** (logcat-only evidence; the debug-only re-arm receiver stays; no production changes; no second microphone).

## Re-arm invariant (final remediation)

The re-arm verification now enforces:

> WakeWordDiag DEBUG verified → new `DETECTOR_GENERATION_STARTED` journal event observed → trial may begin

- A **pre-rearm journal boundary** (sequence) and the **prior authoritative started generation** are captured before the re-arm broadcast.
- Only a strictly post-boundary `DETECTOR_GENERATION_STARTED` with a positive generation **newer than the prior started generation** satisfies the wait (existing ~20 s bounded poll, fail closed).
- `DETECTOR_REARMED` or any other event merely carrying a newer generation is **never** accepted as startup proof — the service journals REARMED before asynchronous detector initialisation emits GENERATION_STARTED, so the old check could declare success before the generation existed.
- New module helper `latest_started_generation(events, after_sequence)` enforces both the event-type and boundary conditions; `_read_target_journal()` reuses the existing GET_JOURNAL_SEQUENCE/GET_JOURNAL_SNAPSHOT interfaces.

## Retained (unchanged since prior review)

- Per-trial energy evidence from the existing #1441 WakeWordDiag `gateExitSummary` logcat lines → `target_timing.gate_episode_energy`, generation-filtered, `none`/missing → `null`, malformed → `gate_episode_energy_error`; private bounded dump at `trials/<trial_id>/target/wakeworddiag.log`, never published.
- `log.tag.WakeWordDiag=DEBUG` set/verified before re-arm; original restored in `cleanup()`; fail closed on set/verify failure.
- Debug-only `WakeDetectorRearmReceiver` (pause/resume of the running service — no new mic, no service restart).
- Source playback metadata (`playback_start_wall_clock_ms`, derived `playback_duration_ms`, applied/requested/max volume, route, focus, fixture hash).
- Correlation via existing `trial_id` + boundary generation.

## Tests (EvidenceRetentionTests, 18)

New for this remediation: `test_rearm_rejects_rearmed_without_generation_started` (the exact regression: newer `DETECTOR_REARMED`, no post-boundary `DETECTOR_GENERATION_STARTED` → fail closed), `test_rearm_rejects_pre_boundary_generation_started` (boundary protection), `test_latest_started_generation_ignores_rearmed_events` (unit-level event-type/boundary filtering). Updated: successful path now requires the post-boundary STARTED event; no-new-generation and rejected-broadcast fail-closed tests retained; property set-before-re-arm, restore-on-cleanup, logcat parsing/filtering, retention/privacy, and no-second-mic guard all green.

## Validation

| Check | Result |
|---|---|
| scripts/tests (749) | OK |
| `:app:testDebugUnitTest` (1077) | OK |
| `assembleDebug` | OK |
| `lint` | 0 errors |
| `git diff --check` | clean |
| Files changed vs prior head | runner + runner tests only |

CI (Build & Test + Docs Drift) green at `af8e1ace`.

No physical test was run. After review/merge, the next step is one bounded 5-trial 120 s S23U→S21 wake-only reproducibility check.

**Do not merge — wait for review.**


NickMonrad (changes requested):
Changes required — one material blocker.

**High — the new target PCM receiver changes the system being measured and cannot prove what the wake detector actually received.**

Location: `app/src/debug/java/com/kernel/ai/debug/capture/TargetCaptureReceiver.kt` (plus its manifest entry and the new main-source journalling changes in `AcousticEvent.kt` / `OnnxWakeWordDetector.kt`).

The accepted #1410 follow-up explicitly bounded this prerequisite to the harness/evidence path, script-only, and allowed a low-bandwidth target energy/level trace instead of raw PCM. This PR instead starts a second concurrent `AudioRecord(VOICE_RECOGNITION, 16 kHz mono int16)` while the production wake detector already owns its own `AudioRecord` with the same source/config. That second recorder is not the detector's PCM stream, and opening another capture client can contend with or alter the microphone capture/preprocessing path on the physical S21. A 5-trial result collected this way therefore cannot safely answer the question we are trying to answer: whether the detector received materially comparable acoustic input to the earlier uninstrumented run. It also contradicts the PR claim that wake behaviour is untouched.

There is a concrete timing problem inside the same design as well: `startCapture()` returns success immediately after starting a thread, before that thread has successfully called `recorder.startRecording()`. Source playback begins after that success, so the capture can miss the phrase onset or fail asynchronously after the runner has already treated capture as started.

**Smallest remediation:** keep this prerequisite script-only as the #1410 decision specified. Remove the second `AudioRecord` receiver and the on-device journalling additions. Persist the already-existing #1441 `WakeWordDiag` gate-exit energy fields per trial from the harness/logcat path, together with the source playback level/route/timing metadata. That satisfies the accepted `raw PCM or at least a low-bandwidth energy/level trace` requirement without perturbing the detector being measured. Preserve the private/sanitised evidence handling and bounded 5-trial follow-up.

Do not broaden this into a detector PCM-tap feature unless the existing detector-derived energy trace is proven insufficient for the bounded A-vs-B decision.

NickMonrad (changes requested):
Focused re-review of the remediation only. The second AudioRecord path is correctly removed and the PR is now runner-only, but one material defect remains.

**High — later runs enable WakeWordDiag after the detector generation has already latched diagnostics off.** `load_later_run_preflight()` calls `ensure_target_diagnostics()` after loading the prior preflight manifest, but it does not re-arm/restart the detector before the first trial. `OnnxWakeWordDetector.runDetectionLoop()` samples `isWakeWordDiagnosticLoggingEnabled()` once when the generation starts and constructs `WakeWordGateExitDiagnostics` only from that latched value. Because preflight cleanup restores `log.tag.WakeWordDiag` to its original value, a subsequent regression/smoke invocation can set the property to DEBUG while the already-running detector generation remains diagnostics-disabled.

Concrete failure: the bounded 5×120s S23U→S21 run starts with a detector generation that began under INFO. The first intended wake misses — exactly the case this measurement is meant to explain — so no activation/re-arm occurs. `gateExitSummary` is therefore absent for that miss, and potentially for subsequent misses in the same generation, leaving `gate_episode_energy` empty and defeating the stimulus-equivalence measurement.

Smallest remediation: after setting/verifying WakeWordDiag DEBUG for any later/resumed physical run, force the detector through the existing approved re-arm lifecycle before the first trial and verify that a new detector generation started under DEBUG (using the existing journal/generation evidence). Do not add production code or another microphone path. Add a focused test proving a later run cannot begin trials until the post-DEBUG re-arm/generation transition is confirmed.

NickMonrad (changes requested):
Focused re-review of the remediation at `20ed633dc4d8634fb8a4ff1fa8fca0a071097013` found one remaining material defect.

**High — the re-arm proof can pass before `DETECTOR_GENERATION_STARTED` occurs.**

Location: `scripts/acoustic_wake_reliability_runner.py`, `_current_target_generation()` / `_rearm_target_detector()`.

`_current_target_generation()` resolves the newest positive generation from *any* journal event. But `WakeWordService.rearmDetector()` calls `wakeWordDetector.start(generationId=...)` asynchronously and then immediately records `DETECTOR_REARMED` for that generation. The detector records `DETECTOR_GENERATION_STARTED` only later, after its AudioRecord/ONNX sessions initialise.

Concrete failure: previous generation 4 -> debug receiver pause/resume allocates generation 5 -> service immediately journals `DETECTOR_REARMED(g=5)` -> runner sees generation 5 and returns success -> trial execution is allowed even if the detector thread has not yet reached, or never reaches, `DETECTOR_GENERATION_STARTED(g=5)`. If detector initialisation then fails, the measurement can proceed without the WakeWordDiag generation this prerequisite exists to guarantee.

This does not satisfy the required invariant `DEBUG verified -> new detector generation started -> trial begins`.

Smallest remediation: after the broadcast, wait specifically for an authoritative `DETECTOR_GENERATION_STARTED` event after the pre-rearm journal boundary and require its generation to be newer than the previous generation. Do not accept `DETECTOR_REARMED` or another positive-generation event as proof. Add a regression test where a newer `DETECTOR_REARMED` is present but no newer `DETECTOR_GENERATION_STARTED`; setup must fail closed.

No other material defect found in this remediation.

NickMonrad (approved):
Focused re-review of the remediation at `af8e1ace77cb23c016a04c14f76f1e79eb2272ba` only.

The previously identified re-arm verification defect is resolved. The runner now captures a pre-rearm journal boundary, derives the prior generation only from authoritative `DETECTOR_GENERATION_STARTED` events, and accepts setup only after a strictly post-boundary `DETECTOR_GENERATION_STARTED` for a newer positive generation. `DETECTOR_REARMED` can no longer satisfy the proof early.

The added regression coverage exercises the exact failure path (newer `DETECTOR_REARMED` without a started generation), rejects pre-boundary started events, and verifies the successful post-boundary started path. No new material defect is introduced by this remediation.

CI and Docs Drift are green at this exact head. PR #1448 is ready to merge. After merge, proceed with the bounded 5-trial 120-second S23U→S21 wake-only reproducibility check; do not run another full #1410 matrix yet.

PR #1460, opened 2026-08-11
fix(#1459): hydrate chat model settings when attaching to warm engine

### LINKED ISSUE #1459: Chat model settings unavailable when ViewModel attaches to an already-ready inference engine
## Summary

The in-chat **Model settings** sheet can incorrectly report:

> Model settings are available once the chat model is ready.

while the chat/inference engine is already ready and usable.

Observed on the **Honor Magic 8 Pro** during normal app use on 11 Aug 2026. The chat screen was fully available, but tapping the tune/sliders icon opened only the unavailable-state sheet instead of the temperature/top-p/top-k/thinking controls.

This appears to be a lifecycle/state hydration regression in the #961 implementation rather than an Honor-specific inference failure.

## Observed behaviour

1. Open Jandal with the conversation model already loaded/warm.
2. Enter/open a chat backed by a newly-created `ChatViewModel`.
3. Tap the tune/sliders **Model settings** icon.
4. The sheet opens, but displays `Model settings are available once the chat model is ready.`
5. The normal chat surface is already ready, so the message is incorrect and the controls are inaccessible.

## Expected behaviour

When the singleton inference engine is already ready, a newly-created `ChatViewModel` must hydrate its local active-model/settings state from the selected conversation model so the in-chat settings sheet renders the real controls.

No model reload should be required merely to make the settings UI available.

## Root cause assessment

`InferenceEngine` is bound as a singleton, while `ChatViewModel` owns local model metadata such as:

- `activeModel`
- `activeModelState` / `currentModel`
- `_activeModelSettings`
- `_showThinkingProcess`
- active context-window/settings state

Both `initEngineWhenReady()` and `initGemma4()` currently return immediately when the singleton engine is already ready:

```kotlin
if (inferenceEngine.isReady.value) return

val preferred = downloadManager.preferredConversationModel()
val modelPath = downloadManager.getModelPath(preferred) ?: return
activeModel = preferred
activeModelState.value = preferred
```

That means a fresh `ChatViewModel` can attach to an already-warm engine without ever populating its own `activeModelState` or settings snapshot.

`ChatUiState.Ready` can still be reached because the engine/conversation are ready, but the settings sheet calculates readiness from `currentModel != null && modelCapabilities != null`. Because those ViewModel-local values remain null, the sheet incorrectly falls back to its unavailable state.

This is a credible shared lifecycle path, not just a device-specific rendering problem.

## Required change

Keep this fix narrowly scoped to restoring ViewModel state when the engine is already warm.

1. In `initEngineWhenReady()`, do not return from the already-ready path before hydrating the active model/settings state required by `ChatViewModel`.
2. Apply the equivalent correction to `initGemma4()` because it has the same early-return pattern.
3. Reuse the existing preferred conversation model and persisted settings; do not introduce a second model-selection mechanism or reload the engine.
4. Populate only the same local state that a successful normal engine initialisation would leave authoritative for chat/model-settings UI.
5. Do not redesign the inference lifecycle, model settings architecture, or #961 apply/reconfigure path.

## Acceptance criteria

- [ ] A new `ChatViewModel` attached while `InferenceEngine.isReady == true` hydrates `currentModel` to the preferred loaded conversation model.
- [ ] `activeModelSettings` is populated from persisted settings for that model.
- [ ] `ChatUiState.Ready.modelCapabilities` is non-null for the loaded model where capabilities are defined.
- [ ] Tapping the in-chat model-settings icon while the engine is already warm shows the actual supported settings controls rather than the false `model is ready` unavailable message.
- [ ] The fix does not reload/reinitialise an already-ready inference engine.
- [ ] Existing cold-start/model-initialisation behaviour remains unchanged.
- [ ] Regression coverage explicitly starts the fake/mock inference engine in the ready state before constructing/initialising `ChatViewModel` and verifies model/settings hydration.
- [ ] Cover both `initEngineWhenReady()` and `initGemma4()` shared early-return behaviour, either directly or through a common helper if the smallest implementation naturally supports that.
- [ ] `./gradlew :feature:chat:testDebugUnitTest` passes.
- [ ] `./gradlew assembleDebug` passes.
- [ ] Device smoke test verifies the reported flow on Honor Magic 8 Pro; another device is optional unless the implementation exposes a device-specific difference.

## Scope

This is a small regression fix. Do not:

- change sampler semantics;
- change settings defaults;
- add engine-scoped live reload;
- refactor the singleton engine lifecycle;
- reopen or broaden #961.

## Related

- #961 — Chat UX: make in-chat model settings apply without app restart
- PR #1252 — original #961 implementation
- PR #1253 / #1254 — corrective #961 fixes
- #963 — Chat UX overhaul
- #1329 — v0.1 release-candidate golden-journey validation (Honor Magic 8 Pro is the primary validation device)

lokhor:
Closes #1459

## Confirmed root cause

`InferenceEngine` is a process singleton, but `ChatViewModel` owns local model metadata (`activeModel`, `activeModelState`/`currentModel`, `_activeModelSettings`, `_showThinkingProcess`, `activeContextWindowSize`). Both `initEngineWhenReady()` and `initGemma4()` returned immediately when `inferenceEngine.isReady.value` was already true:

```kotlin
if (inferenceEngine.isReady.value) return
```

A fresh `ChatViewModel` created after the engine had already warmed up (previous ViewModel, or the app resumed into chat) therefore never hydrated its local state. `ChatUiState.Ready` was still reached (engine + conversation ready), but the in-chat settings sheet derives readiness from `currentModel != null && modelCapabilities != null` — both null — so it falsely showed "Model settings are available once the chat model is ready." while chat was fully usable.

Reproduced in unit tests (warm engine before ViewModel init → `currentModel`/`activeModelSettings` stay null) and confirmed the sheet shows the unavailable state when hydration is skipped.

## Changes

| File | Change |
|---|---|
| `feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt` | New shared `hydrateActiveModelState()` (resolves preferred conversation model + persisted `ModelSettingsEntity`, populates `activeModel`, `currentModel`, `_activeModelSettings`, `_showThinkingProcess`, `activeContextWindowSize`). The already-ready branches of both `initEngineWhenReady()` and `initGemma4()` now hydrate before returning; the cold-start paths reuse the same helper (same assignments, same `ModelConfig`, same post-initialize `resolvedMaxTokens` sync — no behaviour change). |
| `feature/chat/src/test/java/com/kernel/ai/feature/chat/ChatViewModelInitTest.kt` | 2 regression tests (see below). |
| `feature/chat/src/test/java/com/kernel/ai/feature/chat/ChatViewModelSettingsApplyTest.kt` | `apply without active model sets error` previously relied on the bug (warm engine ⇒ never hydrated). Now stubs `getModelPath → null` so the ViewModel genuinely has no active model; the tested contract (apply errors without persisting/reconfiguring) is unchanged. |

## Why the warm engine is never reloaded

The already-ready branch calls only `hydrateActiveModelState()` (local state assignment + `downloadManager`/`modelSettingsRepository` reads) and syncs `activeContextWindowSize` from the engine's published `resolvedMaxTokens` (the same read the cold path does after `initialize()`). It never calls `InferenceEngine.initialize(...)`, `resetConversation()`, `reconfigureConversation(...)` or `updateSystemPrompt(...)`. Regression tests assert `initialize` is not called on the warm path.

## Regression tests

Both exercise the real lifecycle (no field-setup shortcuts):

1. `fresh viewmodel on warm engine hydrates model and settings without reinitialising` — engine `isReady == true` **before** ViewModel construction; preferred model already downloaded; verifies `currentModel == GEMMA_4_E4B`, `activeModelSettings == persisted settings`, `ChatUiState.Ready.modelCapabilities` non-null, sampler/thinking values from persisted settings, and `initialize` never called. **Fails on pre-fix code** (verified by running it against the reverted implementation).
2. `sendMessage waiting on warmup mutex hydrates state without double initialising` — covers `initGemma4`'s already-ready branch via the real race: `initEngineWhenReady` holds the init mutex while `sendMessage`'s `initGemma4` waits; engine flips ready while it waits; verifies hydration and `initialize` called exactly once (only by the cold path).

## Build results

- `./gradlew :feature:chat:testDebugUnitTest` — **618 tests, 0 failures**
- `./gradlew assembleDebug` — **BUILD SUCCESSFUL**

## Device evidence (S23U — Honor Magic 8 Pro not connected)

Smoke-tested on the connected S23U (issue's Honor device unavailable; second device is optional per issue):

1. Cold start: LiteRT GPU environment + `Gemma4DataProcessor` created once (17:01), engine warm.
2. Back → re-entered chat: fresh `ChatViewModel` attached to the warm engine.
3. Tapped the Model settings icon: sheet rendered the **real controls** — "Gemma 4 E-2B / GPU / Temperature 1.0 / Top-P 0.95 / Top-K 64 / Thinking mode" — not the false not-ready message.
4. No reload: logcat shows no new LiteRT environment and no `Engine ready — backend:` (initialize's completion marker) after re-entry; only the pre-existing persona-collector `updateSystemPrompt`/conversation-reset.

### DIFF
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/InferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/InferenceEngine.kt
@@ -38,6 +38,15 @@ interface InferenceEngine {
     val resolvedMaxTokens: StateFlow<Int>
 
     /**
+     * Absolute path of the model currently loaded by the engine, or null before
+     * [initialize] completes. Authoritative for clients that attach to an
+     * already-ready singleton engine: the engine keeps running the model it was
+     * initialized with even if the user changed their preferred conversation model
+     * (preference changes take effect on next launch).
+     */
+    val loadedModelPath: String?
+
+    /**
      * Load model weights and initialize the LiteRT-LM engine.
      * Blocks the [LlmDispatcher] thread; observe [isReady] for completion.
      *
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -795,6 +795,9 @@ class LiteRtInferenceEngine @Inject constructor(
     private var conversation: com.google.ai.edge.litertlm.Conversation? = null
     private var currentConfig: ModelConfig? = null
 
+    /** The model currently loaded by the engine, from the resolved [initialize] config. */
+    override val loadedModelPath: String? get() = currentConfig?.modelPath
+
     /** Ensures only one generation (chat or isolated) runs at a time. */
     private val generationMutex = Mutex()
 
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt
@@ -964,9 +964,11 @@ class ChatViewModel @Inject constructor(
     private suspend fun hydrateActiveModelState(): Triple<KernelModel, String, ModelSettingsEntity>? {
         val preferred = downloadManager.preferredConversationModel()
         val modelPath = downloadManager.getModelPath(preferred) ?: return null
-        val settings = modelSettingsRepository.getSettings(preferred.modelId)
+        // Assign the model before reading persisted settings so a settings-read failure
+        // leaves the same partial state the original cold-start path did (review #1460).
         activeModel = preferred
         activeModelState.value = preferred
+        val settings = modelSettingsRepository.getSettings(preferred.modelId)
         activeContextWindowSize = settings.contextWindowSize
         _showThinkingProcess.value = settings.showThinkingProcess
         _activeModelSettings.value = settings
@@ -989,20 +991,20 @@ class ChatViewModel @Inject constructor(
             .first()
 
         gemma4InitMutex.withLock {
-            if (inferenceEngine.isReady.value) {
-                // The engine is a process singleton — a previous ChatViewModel may already
-                // have it warm. Never re-initialise it; just hydrate the ViewModel-local
-                // model/settings state the in-chat settings UI needs (#1459).
-                hydrateActiveModelState()
-                // Mirror the cold path's post-initialize sync: the engine allocated a
-                // clamped KV-cache size that supersedes the persisted setting.
-                inferenceEngine.resolvedMaxTokens.value.takeIf { it > 0 }?.let {
-                    activeContextWindowSize = it
-                }
-                return
-            }
-
             try {
+                if (inferenceEngine.isReady.value) {
+                    // The engine is a process singleton — a previous ChatViewModel may already
+                    // have it warm. Never re-initialise it; just hydrate the ViewModel-local
+                    // model/settings state the in-chat settings UI needs (#1459).
+                    hydrateActiveModelState()
+                    // Mirror the cold path's post-initialize sync: the engine allocated a
+                    // clamped KV-cache size that supersedes the persisted setting.
+                    inferenceEngine.resolvedMaxTokens.value.takeIf { it > 0 }?.let {
+                        activeContextWindowSize = it
+                    }
+                    return
+                }
+
                 val (preferred, modelPath, settings) = hydrateActiveModelState() ?: return
                 // EmbeddingGemma uses CPU only (no GPU conflict with Gemma-4).
                 // embeddingEngine.close() removed — it silently broke search_memory (#445)
@@ -1051,18 +1053,18 @@ class ChatViewModel @Inject constructor(
      */
     private suspend fun initGemma4() {
         gemma4InitMutex.withLock {
-            if (inferenceEngine.isReady.value) {
-                // Same already-ready hydration as [initEngineWhenReady] (#1459):
-                // a fresh ViewModel may attach to a warm singleton engine — hydrate
-                // local state only, never re-initialise.
-                hydrateActiveModelState()
-                inferenceEngine.resolvedMaxTokens.value.takeIf { it > 0 }?.let {
-                    activeContextWindowSize = it
-                }
-                return
-            }
-
             try {
+                if (inferenceEngine.isReady.value) {
+                    // Same already-ready hydration as [initEngineWhenReady] (#1459):
+                    // a fresh ViewModel may attach to a warm singleton engine — hydrate
+                    // local state only, never re-initialise.
+                    hydrateActiveModelState()
+                    inferenceEngine.resolvedMaxTokens.value.takeIf { it > 0 }?.let {
+                        activeContextWindowSize = it
+                    }
+                    return
+                }
+
                 val (preferred, modelPath, settings) = hydrateActiveModelState() ?: return
                 // EmbeddingGemma uses CPU only (no GPU conflict with Gemma-4).
                 // embeddingEngine.close() removed — it silently broke search_memory (#445)
--- a/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt
+++ b/feature/chat/src/main/java/com/kernel/ai/feature/chat/ChatViewModel.kt
@@ -948,6 +948,32 @@ class ChatViewModel @Inject constructor(
     }
 
     /**
+     * Resolves the preferred conversation model and hydrates the ViewModel-local
+     * active model/settings state required by the in-chat model settings UI.
+     *
+     * Shared by the cold-start init paths and the already-ready path: the
+     * [InferenceEngine] is a process singleton, so a fresh [ChatViewModel] can attach
+     * while the engine is already warm. In that case the engine must NOT be
+     * re-initialised, but `currentModel`/`activeModelSettings` (and thus
+     * `ChatUiState.Ready.modelCapabilities`) must still be populated, or the in-chat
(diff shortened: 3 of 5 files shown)


lokhor:
## S23U on-device verification (2026-08-12, Wi-Fi ADB, SM-S918B)

Installed the fix build (commit `5b8e04c1`, versionName 0.1.0) via `adb install -r`.

**1. Cold start (logcat cleared at launch):**
- `09:18:22 Trying backend: GPU`
- `09:18:37 Backend GPU initialized successfully`
- `09:18:39 Engine ready — backend: GPU, maxTokens: 8000` ← engine warmed exactly once

**2. Fresh ViewModel on the warm engine (logcat cleared, then BACK → re-enter chat):**
- `09:19:10 Restored session …` (new ChatViewModel attached)
- `09:19:11 System prompt updated and conversation reset` (pre-existing persona-collector attach behavior)

**3. Model settings sheet after tapping the tune icon:** real controls rendered — `Gemma 4 E-4B / GPU / Temperature 1.0 / Top-P 0.95 / Top-K 64 / Thinking mode` — NOT the false `Model settings are available once the chat model is ready.`

**4. No reload on re-entry/sheet:** after the re-entry logcat clear there is no `Trying backend`, no `Backend … initialized`, no `Engine ready`, no `Creating LiteRT environment`. The single `Creating Gemma4DataProcessor` at 09:19:10 is the audio pipeline (immediately followed by a `mel_filterbank.cc` warning), not the conversation engine.

The reported Honor scenario (warm engine + fresh ViewModel → false not-ready sheet) is fixed on-device.

lokhor:
## Honor Magic 8 Pro verification (user device test, 2026-08-12)

Tested by the reporter on the primary device from #1459. Reported outcome:

1. **Model settings sheet shows the real controls** — temperature, top-p, top-k, and thinking mode — after a fresh `ChatViewModel` attached to the already-warm conversation engine. The false `Model settings are available once the chat model is ready.` state no longer appears.
2. **No visible model reload** when entering the new chat or opening the settings sheet.

Full evidence chain for #1459 is now: warm-engine unit regression (fails pre-fix) + S23U ADB run (logcat-verified, no initialize markers) + Honor Magic 8 Pro on-device confirmation.

lokhor:
## Response to review

Reviewer finding (settings-read failure semantics) fixed in `881f8bea`:

1. **Warm branches now run inside the init try/catch** in both `initEngineWhenReady()` and `initGemma4()` — a `getSettings()` failure is logged and surfaced via `_error` (`Failed to load AI model`) exactly like the cold path, instead of escaping the viewModelScope coroutine.
2. **`hydrateActiveModelState()` assigns `activeModel`/`activeModelState` before the settings read**, restoring the original cold-start partial-state order on settings-read failure.
3. New regression test `warm engine settings read failure surfaces error without reinitialising` — warm engine + settings read failure ⇒ `ChatUiState.Ready.error` contains `Failed to load AI model`, `initialize` never called.

Re-verified: 619/619 `:feature:chat:testDebugUnitTest`, `assembleDebug` green; the warm-engine hydration test still fails on pre-fix `origin/main` code.

lokhor:
Independent reviewer subagent verdict on `881f8bea`:

> The prior finding is fixed: both ready branches hydrate inside cancellation-aware try/catch, model identity is assigned before the settings read, and the new regression test covers the failure without initialization. No remaining blocking or non-blocking correctness findings; **APPROVE**.

lokhor:
## Remediation for review findings (commit `522a7527`)

### Finding 1 — warm hydration could identify the wrong loaded model

Added the minimum read-only identity to the engine:

- `InferenceEngine.loadedModelPath: String?` — absolute path of the model currently loaded, backed by `LiteRtInferenceEngine.currentConfig.modelPath` (set in `initialize()`, cleared only on shutdown/release which also clear `isReady`).
- `ChatViewModel.loadedConversationModel()` resolves the loaded `KernelModel` from that path.
- Both already-ready branches (`initEngineWhenReady()` and `initGemma4()`) now hydrate from **the loaded model**, never from `preferredConversationModel()`. Cold initialization still uses the preferred model, unchanged.

Regression test `warm engine hydrates the loaded model not the changed preference`: engine ready + loaded E4B, preference changed to E2B, fresh ViewModel ⇒ `currentModel == GEMMA_4_E4B`, E4B settings hydrated, E2B settings never consulted (`coVerify`), `initialize()` never called. **Verified to fail against the previous fix `881f8bea`** (preference-based hydration).

### Finding 2 — settings-read failure exposed fabricated defaults as real settings

Hydration is now **atomic**: `hydrateActiveModelState(model)` assigns `activeModel` / `activeModelState` / `activeModelSettings` only after `getSettings()` succeeds. A settings-read failure leaves `currentModel` null and `activeModelSettings` null, so the settings sheet's existing readiness gate (`currentModel != null && capabilities != null`) stays closed and Apply fails with `no active model` instead of persisting fallback values. The `Failed to load AI model: ...` error behaviour is unchanged.

Extended regression test `warm engine settings read failure surfaces error and blocks apply`: error surfaced, currentModel/activeModelSettings null, `initialize` never called, and `applyModelSettingsAndStartNewChat(fabricatedDraft)` ⇒ error, `saveSettings` and `reconfigureConversation` never called.

### Validation

- `./gradlew :feature:chat:testDebugUnitTest` — **620 tests, 0 failures**
- `./gradlew :core:inference:testDebugUnitTest` — green (interface change)
- `./gradlew assembleDebug` — **BUILD SUCCESSFUL**
- Warm-engine regression test still fails on pre-fix `origin/main` code (re-verified)
- No device testing performed for this remediation (behaviour change is unit-covered; no runtime-behaviour change beyond the two findings)

Issue #1453, opened 2026-08-10
Bug: weather QIR can route successfully while dropping location or forecast semantics


NickMonrad:
## Summary

Weather routing currently has a systemic semantic-integrity gap: a request can correctly route to `get_weather` while the extracted parameters no longer represent what the user asked for.

The two observed on-device failures are examples of the same underlying class rather than isolated phrases:

```text
Heard: What's the weather in Sydney for the next 5 days
Action: get_weather
Result: I couldn't find that location for weather...
```

Expected:

```text
location = Sydney
forecast_days = 5
```

and:

```text
Heard: What's the weather forecast for Sydney
Action: get_weather
Result: current/today conditions only
```

Expected:

```text
location = Sydney
forecast_days = 3
```

STT is correct in both cases. The defect is weather query semantics between QIR matching and `GetWeatherSkill` execution.

### Weather skill contract

`GetWeatherSkill` has a simple parameter contract:

- no `forecast_days`, no `day` -> current conditions;
- `forecast_days=1..7` -> multi-day forecast;
- `day=tomorrow` -> tomorrow-only forecast;
- `location` absent -> device location;
- `location` present -> named-location lookup.

Therefore a QIR match that omits `forecast_days` or `day` is not harmless: the skill deterministically changes the request to current conditions.

### Root cause class

`QuickIntentRouter` contains many overlapping weather regexes. Each matcher independently returns its final parameter map. There is no shared weather-semantic normalisation after a weather pattern matches.

This creates several ways to route successfully but lose meaning:

1. a generic location matcher can consume temporal wording into the location;
2. a matcher can recognise the word `forecast` but return only `location` or an empty map;
3. a matcher can preserve `raw_query`, but `GetWeatherSkill` does not consume `raw_query`;
4. metric-specific weather matchers can recognise UV/AQI/sunrise-style questions without preserving an explicitly named location;
5. tests can pass by asserting only `get_weather` even when the parameter map is wrong.

### Confirmed / directly implied same-class gaps

The existing code shows these additional cases need coverage as part of the fix:

#### Bare/unspecified forecast

```text
What's the forecast?
Weather forecast
```

Existing bare/local forecast patterns return an empty parameter map, which means current conditions. An explicit forecast with no duration should use the established 3-day default.

#### Named forecast without duration

```text
What's the weather forecast for Sydney?
What's the forecast for Sydney?
Weather forecast for Sydney
```

These must produce:

```text
location = Sydney
[...]

Issue #1449, opened 2026-08-09
Crash: Kokoro cold-start can terminate Jandal during spoken reply on Honor Magic 8 Pro


NickMonrad:
Parent release tracker: #1014
Related TTS launch gate: #1447
Related Kokoro quality/performance investigation: #854
Diagnostics prerequisite: #1007 (completed)

## Problem

A physical-device crash was observed on the Honor Magic 8 Pro while Jandal was answering a simple date question and beginning Kokoro TTS.

Observed on 2026-08-09 around 13:05 local time:

- app version: `0.1.0` (`versionCode 2916`);
- build commit: `51c0a3f5`;
- device: Honor Magic 8 Pro;
- Kokoro was the selected speech-output engine;
- the user asked Jandal a date question;
- the response had reached the point where Kokoro was starting/synthesising speech;
- the Jandal process terminated;
- no in-app error message was displayed;
- reopening Jandal worked normally.

The original post-crash export was not diagnostically useful because the old exporter only captured current-process PID-scoped logcat. #1007 has since been completed and the improved export now includes historical `ApplicationExitInfo` records.

## Evidence update — 2026-08-11

A diagnostic export from current `main` (`0.1.0`, versionCode `3054`, commit `0d3b8a5c`) recovered the historical Android exit records from 2026-08-09.

The record that aligns with the reported ~13:05 AEST Kokoro incident is:

- timestamp: `2026-08-09T03:04:51.345Z` / approximately `13:04:51 AEST`;
- reason: `REASON_CRASH (4)`;
- importance: `IMPORTANCE_FOREGROUND (100)`;
- description: `crash`;
- RSS: `433148 KB`;
- system trace available: no.

This classifies the reported incident as Android `REASON_CRASH`, not `REASON_CRASH_NATIVE` and not a low-memory process kill. The next investigation should therefore prioritise capturing the uncaught Java/Kotlin/application-level exception and stack at the Kokoro cold-start boundary before changing architecture or adding memory mitigations.

The same history also contains two separate records that must not be conflated with the reported 13:05 incident:

- `2026-08-09T02:39:00.125Z` — another foreground `REASON_CRASH`, RSS `1584276 KB`; potentially relevant supporting evidence, but not yet correlated to Kokoro.
[...]

Issue #1007, opened 2026-05-29
About screen: make exported logs useful for crash diagnostics


lokhor:
Parent release tracker: #1014
Related crash: #1449

## Problem

Settings → About currently exposes **Verbose logging** and **Export logs**, but the exported file is not reliable enough for real crash diagnosis.

The current implementation in `AboutViewModel`:

- runs `logcat -d -v threadtime --pid=<current process pid> -t 500`;
- therefore captures only the currently running Jandal process;
- treats any non-blank stdout as useful logcat;
- fails the whole export when current-process logcat is blank.

That means a user who reopens Jandal after a crash cannot export evidence from the process that actually crashed.

## 2026-08-09 Honor Magic 8 Pro incident

A concrete failure was observed while investigating #1449:

- Jandal `0.1.0` (`versionCode 2916`), commit `51c0a3f5`;
- Honor Magic 8 Pro;
- Jandal terminated while Kokoro TTS was starting for a normal spoken reply;
- no in-app error was shown;
- the app was reopened immediately and **Export logs** was used at approximately 13:05.

The resulting file contained only a short opaque MagicOS/vendor `HKS…HKE` blob. It contained no useful Jandal timestamps, exception, Kokoro/Sherpa diagnostics or previous-process exit information.

Because the exporter filtered by the new process PID, ordinary current-process logcat cannot diagnose the previous process after relaunch.

## Goal

Make **Export logs** produce one bounded, human-readable diagnostic text bundle that remains useful immediately after Jandal has crashed and been reopened.

Prefer the smallest Android-native change. Do not build a logging framework, crash-reporting service, tombstone decoder or persistent log store.

### 1. Include recent Jandal process-exit information

Use `ActivityManager.getHistoricalProcessExitReasons(...)` / `ApplicationExitInfo` for Jandal's own package.

Include a small fixed number of the most recent records, newest first. For each record include safe diagnostic metadata already exposed by Android, including where available:

- timestamp;
- reason;
- status/signal value;
- process importance;
- description;
- PSS;
- RSS;
- whether a system trace stream is available.

This is the primary path for diagnosing a crash after the process has restarted. Do not infer the previous PID and scrape stale logcat as the crash-history mechanism.

Jandal's current `minSdk` is 35, so no legacy compatibility path is required for `ApplicationExitInfo`.

### 2. Handle `ApplicationExitInfo` traces according to their real format

Do not assume every `traceInputStream` is UTF-8 text.

[...]

PR #1462, opened 2026-08-13
feat(#1461): preserve last uncaught managed exception for post-restart export

### LINKED ISSUE #1461: Debug diagnostics: preserve last uncaught managed exception for post-restart export
Parent crash investigation: #1449
Related diagnostics work: #1007 (completed)
Release tracker: #1014

## Problem

#1007 successfully made post-restart diagnostics useful enough to classify the original Honor Magic 8 Pro Kokoro incident as foreground `ApplicationExitInfo.REASON_CRASH (4)`.

However, the remaining #1449 investigation is now blocked on the concrete Java/Kotlin exception stack:

- the Honor Magic 8 Pro / MagicOS build suppresses usable app/system/crash logcat for both in-app export and ADB;
- the recovered `ApplicationExitInfo` record has no system trace stream;
- DropBox did not retain an app-crash stack for the bounded reproduction session;
- repeated current-main Kokoro cold starts did not reproduce the crash, so there is no stack to identify the first failing Jandal/Kokoro boundary.

Without a retained managed exception stack, another intermittent `REASON_CRASH` on this device would still tell us only that the process crashed, not where or why.

## Goal

Add the smallest debug-build-only mechanism needed to preserve the most recent uncaught Java/Kotlin exception and bounded stack trace across process restart, then include it in the existing Settings → About diagnostic export.

This is a diagnostic prerequisite for #1449. It is not a crash-reporting framework and must not change normal crash semantics.

### 1. Capture the last uncaught managed exception in debug builds

During application startup in debuggable builds only:

- retain the process's existing/default `Thread.UncaughtExceptionHandler`;
- install a thin wrapper handler;
- when invoked, synchronously persist a bounded record containing at minimum:
  - UTC timestamp;
  - thread name;
  - throwable class;
  - throwable message when present;
  - bounded stack trace including causes/suppressed detail as produced by the normal throwable stack representation;
- after the best-effort local write, always delegate to the previously installed/default handler so Android still terminates and records the crash normally.

Do not swallow, recover from, restart after, or otherwise alter the uncaught exception.

### 2. Keep persistence minimal and local

Use the simplest existing app-local storage appropriate for one small crash record. A single private file is sufficient.

- Keep only the latest uncaught managed crash record.
- Bound the stored/exported size.
- The handler must tolerate write failure and still delegate to the original crash handler.
- Do not add Room entities, databases, telemetry SDKs, remote upload, rolling logs, or a generic logging abstraction.
- Do not deliberately persist conversation text, prompts, memories, contacts, model output, or other app data beyond what Android/JVM naturally includes in the throwable message/stack.

### 3. Include the retained record in the existing About export

Extend the existing diagnostic bundle from #1007 with a clearly separated section such as:

```text
Last uncaught managed exception
-------------------------------
Timestamp: ...
Thread: ...
Exception: ...
Message: ...
Stack trace:
...
```

If no retained record exists, say so briefly or omit the body; do not fail the export.

Keep the existing sections and `ApplicationExitInfo` behaviour unchanged.

### 4. Avoid stale-crash misclassification

The export must make it possible to determine whether the retained stack plausibly corresponds to a recent `ApplicationExitInfo.REASON_CRASH` record by timestamp.

Do not automatically claim that an old retained stack caused a newer exit. Preserve the timestamp and let #1449 correlate the evidence.

A successful newer capture may replace the older record. No history UI or multi-record store is required.

## Tests

Add focused tests around the new helper/storage/export behaviour. At minimum cover:

- uncaught throwable record formatting includes timestamp, thread, exception class/message and stack;
- record size is bounded;
- persisted record survives a simulated process restart/read path;
- write failure does not prevent delegation to the previous handler;
- previous handler is invoked exactly once with the original thread/throwable;
- release/non-debuggable configuration does not install the debug capture handler;
- About export includes the retained record when present;
- About export still succeeds when the retained record is absent/unreadable and other #1007 evidence exists.

Do not deliberately crash the normal unit-test process.

## Acceptance criteria

- [ ] Debug builds preserve the latest uncaught managed exception stack locally across process restart.
- [ ] The record includes timestamp, thread, throwable class/message and a bounded stack trace.
- [ ] The capture path delegates to the pre-existing/default uncaught exception handler and does not change normal Android crash termination semantics.
- [ ] Capture/write failures cannot suppress or replace the original crash handling.
- [ ] The existing Settings → About diagnostic export includes the retained managed-crash record when available.
- [ ] The retained stack can be correlated manually with `ApplicationExitInfo` by timestamp without the exporter asserting unsupported causality.
- [ ] Release/non-debuggable builds do not enable this diagnostic capture.
- [ ] No remote crash reporting, logging framework, database, telemetry, or broad diagnostics redesign is introduced.
- [ ] Focused tests cover capture, delegation, bounds, persistence and export integration.
- [ ] #1449 can resume with the next intermittent managed crash producing enough evidence to identify the concrete failing boundary.

## Out of scope

- fixing the Kokoro crash itself;
- catching `OutOfMemoryError`, `RuntimeException`, AudioTrack failures, or other speculative #1449 candidates;
- native tombstone capture/parsing;
- ANR capture beyond the existing #1007 `ApplicationExitInfo` path;
- continuous/persistent application logging;
- remote analytics/crash reporting;
- user-facing crash history UI;
- changes to TTS architecture or fallback behaviour.

lokhor:
Closes #1461

Debug-build-only diagnostic capture: preserve the most recent uncaught Java/Kotlin exception (with bounded stack) across process restart, and include it in the existing Settings → About diagnostic export. Prerequisite for #1449 — the next intermittent managed crash on the Honor Magic 8 Pro (where logcat is suppressed) will leave a concrete stack to identify the failing boundary.

## Where the handler is installed

`KernelAIApplication.onCreate()` (app module), immediately after `super.onCreate()`. Installation is gated on `(applicationInfo.flags and FLAG_DEBUGGABLE) != 0` via `installUncaughtExceptionCaptureIfDebuggable(...)` — **release/non-debuggable builds never install it**. The current/default `Thread.getDefaultUncaughtExceptionHandler()` is retained and wrapped.

## Where the record is stored

One private app-local file: `filesDir/last_uncaught_exception.txt` (single record, latest capture replaces the previous one via overwrite). Format (human-readable, UTC ISO-8601 timestamp for manual correlation with `ApplicationExitInfo`):

```text
Timestamp: 2026-08-13T01:43:27.123Z
Thread: main
Exception: java.lang.RuntimeException
Message: ...
Stack trace:
java.lang.RuntimeException: ...
	at ...
Caused by: ...
```

The stack uses the normal JVM representation (`printStackTrace`), so `Caused by`/`Suppressed` detail is included where naturally produced.

## Bound applied

`LAST_UNCAUGHT_EXCEPTION_MAX_BYTES = 64 KiB` (UTF-8), applied to the complete record; oversized records are truncated on a character boundary with an explicit `[... record truncated ...]` marker. The reader also caps at the same bound.

## How normal uncaught handling is preserved

The wrapper synchronously writes the record best-effort (`runCatching` — a write failure returns false and is ignored), then **always** delegates to the previously installed handler with the original thread and throwable, exactly once. It never swallows, recovers, restarts after, or otherwise alters the crash. On Android the platform default handler is always present, so termination/`REASON_CRASH` recording is unaffected.

## How the About export consumes it

`AboutViewModel.buildDiagnosticExport` (the #1007 bundle) gains a `Last uncaught managed exception` section between *Recent process exits* and *Current process logcat*. Absent → one-line note; unreadable → warning appended to *Exporter warnings*; either way the export succeeds as long as existing #1007 evidence exists. No automatic claim of correspondence to any `ApplicationExitInfo` record — the timestamp is preserved for manual correlation only. All existing #1007 exit/trace/logcat behaviour is unchanged.

## Branch note

The branch was created before #1452 (#1007) reached `main`; it has since been merged with current `main` (which includes #1007 and #1453–#1460), so the PR diff contains **only** the #1461 changes (5 files, all additions).

## Tests executed

- `:feature:settings:testDebugUnitTest` — **164 tests, 0 failures** (full module suite on the merged main).
- New `LastUncaughtExceptionRecordTest` (14 tests): formatting fields (timestamp/thread/class/message/stack incl. caused-by/suppressed); bounded output; write→read persistence across a simulated restart; write failure returns false without throwing; handler delegates exactly once with original thread/throwable; delegation survives persistence failure; null-previous handler; non-debuggable config does not install; debuggable config installs a working handler; absent/unreadable record reads.
- New AboutViewModel export tests: record section included when present; export succeeds with a note when absent; export succeeds with a warning when unreadable.
- `./gradlew assembleDebug` — BUILD SUCCESSFUL on the merged tree.

No deliberate crash is triggered in any test.

### DIFF
--- a/app/src/main/java/com/kernel/ai/KernelAIApplication.kt
+++ b/app/src/main/java/com/kernel/ai/KernelAIApplication.kt
@@ -2,6 +2,7 @@ package com.kernel.ai
 
 import android.app.Application
 import android.content.ComponentCallbacks2
+import android.content.pm.ApplicationInfo
 import android.util.Log
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
@@ -18,6 +19,8 @@ import com.kernel.ai.core.memory.worker.ArchiveCleanupWorker
 import com.kernel.ai.core.memory.worker.MemoryEmbeddingWorker
 import com.kernel.ai.core.memory.worker.WORK_NAME_ARCHIVE_CLEANUP
 import com.kernel.ai.core.memory.worker.WORK_NAME_BACKFILL
+import com.kernel.ai.feature.settings.installUncaughtExceptionCaptureIfDebuggable
+import com.kernel.ai.feature.settings.lastUncaughtExceptionRecordFile
 import dagger.hilt.android.HiltAndroidApp
 import com.kernel.ai.assistant.WakeWordService
 import kotlinx.coroutines.Dispatchers
@@ -49,6 +52,16 @@ class KernelAIApplication : Application(), Configuration.Provider {
 
     override fun onCreate() {
         super.onCreate()
+        // Debug-build-only diagnostic capture (#1461): retain the most recent uncaught
+        // managed exception across process restart for the Settings -> About export.
+        // Never installed in release builds and never alters normal crash handling —
+        // the wrapper delegates to the previously installed handler after the write.
+        installUncaughtExceptionCaptureIfDebuggable(
+            debuggable = (applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE) != 0,
+            recordFile = lastUncaughtExceptionRecordFile(this),
+            currentHandler = Thread.getDefaultUncaughtExceptionHandler(),
+            setHandler = Thread::setDefaultUncaughtExceptionHandler,
+        )
         // Preload ORT so libsherpa-onnx-jni.so (built against ORT as an external shared lib)
         // can resolve OrtGetApiBase at dlopen time. Must happen before any Sherpa TTS or
         // wake word ONNX session init, otherwise OfflineTts.<clinit> fails with UnsatisfiedLinkError.
--- a/feature/settings/src/main/java/com/kernel/ai/feature/settings/AboutScreen.kt
+++ b/feature/settings/src/main/java/com/kernel/ai/feature/settings/AboutScreen.kt
@@ -132,11 +132,23 @@ fun AboutScreen(
             ListItem(
                 modifier = Modifier.fillMaxWidth(),
                 headlineContent = { Text("Export logs") },
-                supportingContent = { Text("Share recent logcat output as a text file") },
+                supportingContent = { Text("Share recent crash information and current logs") },
                 trailingContent = {
                     val isLoading = uiState.exportState is ExportState.Loading
                     Button(
-                        onClick = { if (!isLoading) viewModel.exportLogs() },
+                        onClick = {
+                            if (!isLoading) {
+                                viewModel.exportLogs(
+                                    DiagnosticBuildInfo(
+                                        versionName = versionName,
+                                        versionCode = versionCode,
+                                        buildType = buildType,
+                                        gitSha = gitSha,
+                                        buildTimestamp = buildTimestamp,
+                                    ),
+                                )
+                            }
+                        },
                         enabled = !isLoading,
                     ) {
                         if (isLoading) {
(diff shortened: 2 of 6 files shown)


NickMonrad (approved):
Reviewed against #1461 and the #1449 diagnostic prerequisite. No material defects found. The change is correctly debug-gated, preserves/delegates to the existing uncaught handler, keeps a single bounded app-local record, integrates non-fatally with the existing #1007 export, and does not broaden into crash remediation or TTS changes. CI is green. Ready to merge.

Issue #1418, opened 2026-07-21
Bug: thinking-channel content and protocol tokens leak into the visible assistant reply


NickMonrad:
## Summary

When a Quick Actions request falls through to the LiteRT LLM with thinking mode active, internal reasoning and model protocol tokens can be rendered as user-visible assistant content.

Observed after entering:

```text
Remind me to buy milk
```

The copied transcript contained:

```text
[Thinking]
<|channel>thought
...
[End Thinking]
thought
...
<channel|>Kia ora. When would you like me to remind you to buy milk?
```

The valid answer was eventually produced, but raw channel markers and a later reasoning segment were stored as ordinary visible assistant content.

## Important distinction

Jandal may intentionally show cleaned thinking content in its dedicated collapsible thinking UI when thinking mode is enabled. That is not the defect tracked here.

The `[Thinking]` and `[End Thinking]` lines in the copied transcript are also intentional clipboard annotations when the user chooses to include thinking. They may remain as transcript formatting, but they must never originate from model output or contain raw runtime protocol markers.

The defects are that:

- raw protocol tokens such as `<|channel>thought` and `<channel|>` are exposed;
- reasoning content appears outside the intended thinking presentation;
- a later thought-channel segment is appended to the normal assistant reply;
- leaked content can reach live UI, streaming TTS, persistence, indexing, and export.

## Expected behaviour

For the captured request, the visible assistant reply should contain only:

```text
Kia ora. When would you like me to remind you to buy milk?
```

If thinking mode is enabled, all cleaned reasoning segments may be accumulated in the existing thinking field/presentation. No reasoning or channel syntax may appear in the normal assistant content.

## Confirmed failure area

This is an inference streaming/parser defect, not primarily a Compose or persistence defect.

Automatic LiteRT tool calling can perform multiple model passes inside one `sendMessageAsync()` operation:

```text
model thinks
-> model calls load_skill
-> SDK executes the tool and supplies its result
-> model thinks again
-> model produces the final answer
```

The captured response follows this sequence.

The current `ThinkingStreamStateMachine` makes one irreversible transition from `PRE_CLOSE` to `POST_CLOSE`. After the first close marker, it deliberately treats every later `message.channels["thought"]` delta as visible response text. This is also enforced by the existing `post-close thought channel traffic is emitted as visible response` unit test.

That assumption is invalid for automatic tool-call generations: the first close marker ends one thought segment/model pass, not all thinking for the overall operation.

[...]

Issue #1329, opened 2026-06-27
v0.1 release-candidate golden-journey validation


NickMonrad:
Parent release tracker: #1014
Related publication tracker: #441

## Purpose

Validate the **actual Jandal v0.1 release candidate** through a focused set of end-to-end user journeys on physical devices.

This replaces the previous broad post-#1325 manual matrix. Comprehensive feature permutations remain in living QA issues such as #427 and #824, but they do not all need to pass before v0.1. This issue answers the narrower release question:

> Can a normal user install Jandal, complete setup, use its core advertised capabilities, recover from common Android permission/state changes, and continue without crashes, data loss, or misleading results?

## Build under test

- Release-candidate commit SHA:
- versionName / versionCode:
- Source: local signed build / Play internal track / Play closed track:
- AAB/APK or Play release link:
- Test date:
- Tester:

## Execution strategy

Use the existing automated/ADB/device-test coverage first wherever it exercises the same release behaviour. Do not create a new broad test framework for this issue. Use manual interaction only for journeys or assertions that genuinely require human observation, Android/OEM UI, audio-quality judgement, lock-screen behaviour, or physical-device state changes.

Record which checks were automated versus manually observed so the final release evidence is auditable.

### Primary release validation: Samsung S23 Ultra

Run the complete golden-journey pass on the S23 Ultra where practical. Record:

- Android version/build;
- selected conversation model and backend;
- fresh install vs update install;
- system colours/theme mode;
- Jandal default-assistant and Hey Jandal state.

The S23 Ultra is also the v1 wake-word launch-qualification/reference device, so it is the preferred device for the final Hey Jandal lifecycle journey.

### Compatibility and automated validation: Samsung S21

Use the S21 for as much repeatable validation as the existing harness/device automation supports, plus targeted manual compatibility checks where useful, including:

- clean-install and lower-tier model sanity;
- update/persistence checks where practical;
- deterministic actions such as alarms/timers, weather, lists/notes, memory and permission recovery;
- regression checks already covered by existing automation;
- voice/wake compatibility only as a regression signal, not as a repeat of the completed v1 wake qualification.

Its previously measured wake recall remains non-blocking for v0.1.

### Targeted device-specific validation: Honor Magic 8 Pro

Do **not** duplicate the complete golden-journey matrix on the Honor by default. Use it only where its hardware/OEM behaviour is materially relevant, especially:

- high-memory / larger-model behaviour;
[...]

PR #1522, opened 2026-09-07
fix(1418): filter malformed thinking close marker

lokhor:
## Summary

- treat the exact S21-reproduced `</|/think|>` token as a supported thinking close marker
- retain one incremental parser boundary for split and cumulative callbacks
- add regressions for exact, split, cumulative, repeated, unterminated, normal, and downstream-clean output paths
- filter the separate truncated `|/think>` callback tail observed during bounded S21 smoke without changing the structured callback contract

Structured thought/primary callback handling and downstream chat/TTS/persistence consumers are unchanged.

## Verification

- `:core:inference:testDebugUnitTest --tests com.kernel.ai.core.inference.ThinkingStreamStateMachineTest` — passed (23 tests)
- combined `:core:inference:testDebugUnitTest :feature:chat:testDebugUnitTest :app:testDebugUnitTest :app:assembleDebug` — passed
- current debug APK installed with `adb install -r` on S21 `R5CR605B71K`; latest bounded Hello smoke exercised the app and reached the existing blank-response fallback without protocol-token output. It did not reproduce a normal model response, so Journey 1 is not marked proven.
- J2 S23U baseline was not touched.

Related: #1418, #1329

### DIFF
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -101,6 +101,8 @@ internal fun checkGpuRestartNeeded(
 private const val MIN_AVAIL_MEM_FOR_GPU_BYTES = 2L * 1024 * 1024 * 1024 // 2 GB absolute floor — catches 4-6 GB devices; 8 GB devices pass this
 internal const val THINKING_CHANNEL_HEADER = "<|channel>thought"
 internal const val THINKING_CLOSE_MARKER = "<channel|>"
+/** Exact malformed close marker observed in the S21 Journey 1 regression. */
+internal const val MALFORMED_THINK_CLOSE_MARKER = "</|/think|>"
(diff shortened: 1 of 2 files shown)


NickMonrad (approved):
Reviewed against #1418 and the September 7 #1329 reassessment. No material code defects found in the changed inference parser. The exact malformed S21 marker and observed truncated tail are handled at the existing boundary, with focused split/cumulative, repeated-thinking, non-thinking, and protocol-free output regressions. The change preserves structured callback handling and existing downstream consumers without expanding the architecture.

The implementation is ready to merge once required CI checks pass. The S21 smoke is not a successful Journey 1 validation: it reached the existing blank-response fallback and did not produce a normal model response. Keep #1418 and Journey 1 open until a current-candidate physical-device run demonstrates a complete ordinary local response with no reasoning/protocol leakage. Do not repeat the full golden matrix or disturb the historical Journey 2 baseline. The blank-response observation should be investigated separately if reproducible; do not assume this parser change caused it without evidence.

PR #1523, opened 2026-09-07
fix(1418): contain malformed thinking callbacks

lokhor:
## Summary

Refs #1418 and #1329. The S21 E2B callback stream after PR #1522 emitted a malformed second thought segment:

`<channel|>` → `<|/think>` (split) → second thought text → `</think>` (split) → final answer.

The fallback state machine now recovers that exact sequence only after the observed channel-thought close. This follow-up fixes the remaining P1: if the malformed second-thought candidate never receives its closing `</think>`, the buffered candidate is discarded at completion instead of being restored into visible output.

Smallest change:

- remove unresolved-candidate restoration from `finish()`;
- clear the confirmed private candidate and return to the terminal visible state;
- let the existing blank-response retry/fallback handle a generation with no visible answer;
- replace the prior unterminated-candidate visibility expectation with a fail-closed regression preserving a legitimate visible prefix and verifying the next independent parser remains clean.

No new marker variants, sanitizer, parser layer, ChatViewModel, TTS, Room, RAG, clipboard, QIR, model, or UX changes.

## Verification

- `:core:inference:testDebugUnitTest --tests com.kernel.ai.core.inference.ThinkingStreamStateMachineTest` — PASS
- `:core:inference:testDebugUnitTest :feature:chat:testDebugUnitTest` — PASS
- `:app:assembleDebug` — PASS
- Final debug APK SHA-256: `5b29803f42b5385be0920f47f47a16d0b89ff479710aa1a9ead66265ed1a9398`.
- S21 `R5CR605B71K`, existing E2B cache, no app-data/model-cache clear: final APK installed with `adb install -r`. Physical inference was blocked during this run by the existing debug startup failure `ForegroundServiceStartNotAllowedException: startForegroundService() not allowed due to mAllowStartForeground false` from `InferenceLoadingService`; the app remained on the model-loading screen. No ordinary completion or leak claim is made from this blocked run.
- The established `load_skill`/`run_intent` fixture was attempted on the final APK. The app reached the generation path but ended with `Sorry, generation was cancelled`; the same run logged the foreground-service startup failure. Successful repeated automatic-tool completion is not claimed.

Focused parser coverage proves the unresolved candidate emits no reasoning or protocol fragment, preserves `Visible ` emitted before the confirmed malformed context, and does not affect a subsequent independent generation. #1418, #1329, and Journey 1 remain open. No full golden-journey matrix was restarted.

### DIFF
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -305,8 +305,9 @@ internal class ThinkingStreamStateMachine(
         val response = mutableListOf<String>()
 
         if (rawMode == RawMode.MALFORMED_THOUGHT_CANDIDATE) {
-            rawPending.insert(0, malformedThoughtCandidate.toString())
-            rawPending.insert(0, MALFORMED_THINK_REOPEN_MARKER)
+            // The confirmed malformed-thought candidate is private reasoning. If its
+            // closing marker never arrives, fail closed and let the blank-response
+            // fallback handle the generation instead of exposing it as visible text.
             malformedThoughtCandidate.clear()
             malformedReopenEligible = false
             rawMode = RawMode.VISIBLE
--- a/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
+++ b/core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
@@ -238,6 +238,7 @@ internal class ThinkingStreamStateMachine(
         CHANNEL_HEADER,
         CHANNEL_THOUGHT,
         CHANNEL_VISIBLE,
+        MALFORMED_THOUGHT_CANDIDATE,
     }
 
     private enum class MarkerType {
@@ -261,6 +262,8 @@ internal class ThinkingStreamStateMachine(
     private val emittedThinking = StringBuilder()
     private val emittedStructuredThinking = StringBuilder()
     private val emittedResponse = StringBuilder()
+    private val malformedThoughtCandidate = StringBuilder()
+    private var malformedReopenEligible = false
     private var rawMode = RawMode.VISIBLE
 
     fun consume(channelDelta: String?, rawMessage: String): ThinkingStreamEmission {
@@ -300,6 +303,15 @@ internal class ThinkingStreamStateMachine(
     fun finish(): ThinkingStreamEmission {
         val thinking = mutableListOf<String>()
         val response = mutableListOf<String>()
+
+        if (rawMode == RawMode.MALFORMED_THOUGHT_CANDIDATE) {
+            rawPending.insert(0, malformedThoughtCandidate.toString())
+            rawPending.insert(0, MALFORMED_THINK_REOPEN_MARKER)
+            malformedThoughtCandidate.clear()
+            malformedReopenEligible = false
+            rawMode = RawMode.VISIBLE
+        }
+
         processRaw(thinking, response)
         processStructuredThinking(thinking)
 
@@ -370,6 +382,7 @@ internal class ThinkingStreamStateMachine(
                 RawMode.CHANNEL_HEADER -> processChannelHeader()
                 RawMode.CHANNEL_THOUGHT -> processChannelThought(thinking)
                 RawMode.CHANNEL_VISIBLE -> processChannelVisible(response)
+                RawMode.MALFORMED_THOUGHT_CANDIDATE -> processMalformedThoughtCandidate(thinking)
             }
             if (beforeLength == rawPending.length && beforeMode == rawMode) return
             if (rawPending.isNotEmpty() && rawMode == RawMode.VISIBLE) {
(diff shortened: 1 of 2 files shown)


core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
```diff
@@ -380,6 +384,11 @@ internal class ThinkingStreamStateMachine(
         response: MutableList<String>,
     ) {
         val text = rawPending.toString()
```
NickMonrad, line 400: **P2 — Context-free malformed reopen can suppress valid output.** This branch enters `THOUGHT` for every `<|/think>` in visible content, regardless of whether a preceding thought/channel close established the malformed second-segment sequence. A literal occurrence in an ordinary response can therefore withhold all following text until a close marker (or discard it at finish). The new global `</think>` registration also strips legitimate HTML/XML closing tags from normal output. Limit these recoveries to the observed malformed-thinking context, make the boundary guard agree, and test a literal-markup non-thinking response as well as the captured S21 sequence.

NickMonrad (changes requested):
The captured S21 sequence is now represented by a focused regression fixture, and the two clean device turns are useful progress. I found one material correctness regression that needs remediation before merge.

**P2 — Scope malformed-marker recovery to the observed thinking context.** In `LiteRtInferenceEngine.kt`, the new `processVisible()` branch accepts `<|/think>` whenever it appears, while `MALFORMED_HTML_THINK_CLOSE_MARKER` is registered globally in `THINK_CLOSE_MARKERS` and `PROTOCOL_MARKERS_FOR_BOUNDARY`. A normal non-thinking response containing a literal `<think>example</think>` (for example, an HTML/XML or protocol explanation) now loses the closing tag; a literal `<|/think>` can also switch the parser into thought mode and withhold subsequent valid response text until another close is seen. This changes ordinary visible content and violates #1418's non-thinking-regression and clean-final-response requirements. The physical trace justifies recovering these malformed fragments in its specific second-thought sequence, not treating them as unconditional control syntax everywhere. Restrict the malformed reopen/HTML-close interpretation to the relevant observed parser context, including split-marker buffering, and keep the boundary guard consistent with that classification. Add a regression showing ordinary literal markup survives unchanged, alongside the captured S21 fixture. No new parser layer or broader redesign is needed.

**Remaining acceptance evidence:** After remediation, rerun the focused tests and bounded device check. The current Hello/How smoke does not establish the existing #1418 requirement for repeated automatic-tool generations to finish with a clean user-facing answer. Reuse the established load_skill/run_intent fixture and perform a successful physical multi-pass reproduction where practical; record any genuine device limitation instead of marking the full issue proven from a tool-failure response. Keep #1418 and Journey 1 open until their required evidence is complete. Do not restart the full golden-journey matrix.

core/inference/src/main/java/com/kernel/ai/core/inference/LiteRtInferenceEngine.kt
```diff
@@ -296,6 +303,15 @@ internal class ThinkingStreamStateMachine(
     fun finish(): ThinkingStreamEmission {
         val thinking = mutableListOf<String>()
         val response = mutableListOf<String>()
+
```
NickMonrad, line 306: **P1 — Unterminated candidate leaks reasoning at completion.** This restoration changes a confirmed malformed-thought candidate back into visible text. If the model emits `<|channel>thought\nfirst<channel|><|/think>private reasoning` and then completes without `</think>`, the buffered reasoning can be emitted as `GenerationResult.Token`. The visible guard now permits the literal malformed opener, so it does not prevent this. Keep an entered candidate withheld on incomplete termination and let the existing blank-response fallback handle an absent answer. Update the new incomplete-candidate test to assert no visible reasoning; retain literal preservation for ordinary non-thinking text.

NickMonrad (changes requested):
Re-review of the remediation only. The previous context-free marker finding is substantially resolved: malformed reopening is now gated on the preceding channel-thought close, the HTML close is no longer globally classified as a visible control token, and the literal-markup tests cover the intended preservation behaviour. The focused tests and latest CI pass. One material defect remains in the new candidate-completion path.

**P1 — An unterminated malformed thought is restored as visible output.** In `ThinkingStreamStateMachine.finish()`, the new `MALFORMED_THOUGHT_CANDIDATE` branch prepends the malformed opener and buffered candidate to `rawPending`, switches to `VISIBLE`, and reparses it. For a callback sequence such as `<|channel>thought\nfirst<channel|><|/think>private reasoning` followed by `onDone()` without `</think>`, the buffer is no longer protected as thinking and can be emitted through `GenerationResult.Token`. The boundary guard also now intentionally permits the literal malformed opener, so it cannot distinguish this reasoning from ordinary text. This is a concrete recurrence of #1418's original leak and violates its explicit requirement that unterminated thought must never become visible.

The smallest remediation is to keep an entered malformed-thought candidate fail-closed at completion: withhold/discard the unresolved candidate and use the existing blank-response retry/fallback when no visible answer exists. Preserve already-emitted clean response content. Retain literal-marker preservation outside confirmed malformed-thinking context. Replace the new test that expects an unterminated candidate to become visible with a regression asserting no reasoning or marker reaches any visible delta. No new parser layer or speculative marker handling is needed.

The current two-turn S21 result is useful, but the repeated automatic-tool run ended cancelled after an unknown-skill result. Keep #1418 and Journey 1 open until the existing multi-pass acceptance evidence is satisfied. No full golden-journey rerun is required.

NickMonrad (approved):
Re-review of the remediation at d1306f90c3a9ca2835fb1d4bb30036f6aff05963. Both previous findings are resolved. The malformed reopen remains context-gated, ordinary literal markup is preserved, and finish() now clears an unresolved confirmed thinking candidate rather than restoring it as visible text. The new regression preserves the legitimate visible prefix, asserts that private reasoning and protocol fragments are withheld, and verifies a subsequent independent parser. The change is limited to the existing state machine and focused tests; no new material defects found. Latest Build & Test and docs-drift checks pass. Approved and ready to merge. The final S21 run was blocked by an InferenceLoadingService foreground-service startup failure, so it does not establish physical multi-pass completion. Keep #1418 and Journey 1 open until the existing acceptance evidence is satisfied; investigate that startup failure separately without expanding this parser PR. No full golden-journey rerun is required.

Issue #1261, opened 2026-06-15
Play Store app content declarations: rating, audience, ads, permissions and reviewer access


NickMonrad:
Parent epic: #1014
Related: #441, #1255, #1140

## Purpose

Complete the Play Console App content declarations required before review.

## Current Google Play requirements to reflect

The Play Console App content page is where developers provide policy declarations and legal/safety information. It includes privacy policy, ads, reviewer access instructions, target audience/content, sensitive permission use, content ratings, and privacy/security practices.

References:

- Prepare app for review: https://support.google.com/googleplay/android-developer/answer/9859455
- Content rating requirements: https://support.google.com/googleplay/android-developer/answer/9859655

## Scope

- Complete content rating questionnaire based on actual launch feature set.
- Declare target audience and confirm whether the app is not designed for children.
- Declare whether the app contains ads.
- Provide reviewer access instructions for any gated areas, Hugging Face sign-in/model setup, permission-gated flows, or first-run setup required for review.
- Document and justify sensitive/high-risk permissions and special accesses used by launch scope.
- Ensure permission explanations line up with #1140 permission UX and Play policy declarations.
- Confirm app category/tags and app content settings.

## Acceptance criteria

- [ ] Content rating questionnaire completed and certificate/result recorded.
- [ ] Target audience/content declaration completed.
- [ ] Ads declaration completed.
- [ ] Reviewer access instructions exist and cover model setup/gated downloads/permissions.
- [ ] Sensitive permission and special-access declarations match the release manifest and launch UX.
- [ ] Any Play Console policy warnings from App content are captured and triaged.
- [ ] #1255/#1014/#441 are updated with final App content status.

lokhor:
### Special access to declare in App content: Usage Access — #1502

`android.permission.PACKAGE_USAGE_STATS` is now in the release manifest for the Honor camera-coexistence fix (#1502). It is requested as a **special access** (AppOps), declared and justified only for that use:

- **Scope:** HONOR + model BKQ-N49, the only device where the microphone contention with the stock Honor Camera (`com.hihonor.camera`) was physically proven. All other devices keep the existing wake path with no special access, no prompt, and no foreground polling.
- **Why:** MagicOS refuses stock camera video recording while a background app holds a voice-recognition capture. Jandal watches for the camera entering the foreground so it can release the microphone before recording starts, then re-arms afterwards.
[...]

Issue #1260, opened 2026-06-15
Privacy policy, Data Safety and in-app privacy surface


NickMonrad:
Parent release tracker: #1014
Related: #441, #1261, #1263, #868

## Purpose

Prepare the hosted Jandal privacy policy and Google Play Data Safety declaration for the first Play Store release, and make privacy information clearly accessible from inside the release build.

This issue owns the shared **Settings → About → Legal & Privacy** surface and its privacy-related entries. #1263 owns the open-source licence, model, voice, STT/TTS, and wake-word attribution content presented through the same surface.

## Required v0.1 user experience

The release build must contain a discoverable **Legal & Privacy** area under Settings/About. It must not require the user to find GitHub or the Play Store listing.

At minimum it must provide:

1. **Privacy Policy**
   - Opens the stable hosted Jandal privacy policy in a browser or supported in-app web surface.
   - The hosted policy is the authoritative, updateable version.
   - The link must remain usable if the device is offline, by showing an accurate offline/error state rather than a blank page or silent failure.

2. **Data & Privacy Summary**
   - A concise in-app summary written for normal users.
   - Explains which processing is local and which features can make network requests.
   - Covers model/voice downloads, Hugging Face access where applicable, weather, Wikipedia, currency and other release-exposed network-backed skills.
   - Explains that permissions such as microphone, contacts, calendar and location are used only for the corresponding user-invoked capabilities.
   - Explains where conversations, memories, lists, notes and settings are stored for the launch build, including any backup/export/deletion limitations.
   - Links to the full hosted privacy policy.

3. **Legal and attribution navigation**
   - Provides entries for **Open-source licences** and **Models, voices & wake word**.
   - Their content and release sign-off are owned by #1263.

4. **App identity**
   - Shows app name, version name, version code and, where practical, release/build identifier.

The surface should use normal Jandal navigation and typography, support light/dark/system colours, be accessible with TalkBack, and avoid loading large generated legal content into memory unnecessarily.

## Google Play and privacy scope

Google requires the Data Safety form for apps on closed, open and production tracks. Even apps that do not collect user data must complete the form and provide a privacy policy URL. The form and policy must accurately reflect the actual release artefact and third-party SDK/library behaviour.

References:

- Data Safety form: https://support.google.com/googleplay/android-developer/answer/10787469
- Prepare app for review / privacy policy: https://support.google.com/googleplay/android-developer/answer/9859455

## Scope

[...]

PR #1526, opened 2026-09-11
fix(#1502): yield Hey Jandal while Honor Camera is foreground

### LINKED ISSUE #1502: Bug: Hey Jandal background microphone prevents camera video recording
Parent wake-word epic: #65
Release tracker: #1014
Related wake-word optimisation: #1395 / #1398
Release-candidate journeys: #1329

## Problem

With **Listen for Hey Jandal** enabled on the Honor Magic 8 Pro, the stock Camera app cannot start video recording because Jandal's background wake-word service is already holding the microphone.

Observed on 2026-08-29 on the Honor Magic 8 Pro:

1. Enable **Hey Jandal**.
2. Open the stock Honor Camera app.
3. Switch to **Video**.
4. Attempt to record.
5. Camera refuses to start recording and shows:

> A recording app is already running in the background (Recorder, Call). To record a video, close the app first.

Disabling Hey Jandal removes the conflict.

This is a user-visible device integration defect: enabling Jandal's optional always-listening feature must not make normal camera video recording unusable.

## Current architecture context

The wake-word implementation intentionally keeps a 16 kHz `AudioRecord` open while Hey Jandal is enabled. Existing Jandal-internal voice transitions already stop/release the wake detector before another Jandal STT session takes microphone ownership.

However, there is no equivalent handling for an **external application** such as Camera taking microphone ownership.

#1395/#1398 track battery and inference-duty-cycle optimisation. They acknowledge continuous `AudioRecord` ownership, but they do not own external microphone contention. Keep this issue focused on coexistence with other recording applications.

## Goal

Allow normal external microphone recording, beginning with the stock Camera video path, while Hey Jandal is enabled.

Jandal should relinquish or suspend wake-word microphone capture when required by Android/OEM audio arbitration and automatically re-arm afterward without requiring the user to toggle Hey Jandal off and back on.

## Required investigation

1. Reproduce on current `main` on the Honor Magic 8 Pro with the stock Camera app.
2. Capture relevant logcat/audio-manager state while Camera attempts to begin recording.
3. Determine the actual MagicOS arbitration behaviour:
   - whether Jandal's existing `AudioRecord` is marked silenced when Camera requests the microphone;
   - whether Camera is rejected before Jandal receives any silencing/configuration callback;
   - whether foreground/background recording priority or audio-source selection changes the behaviour.
4. Inspect the existing `WakeWordService` / `WakeWordDetector` microphone lifecycle and reuse the smallest existing stop/re-arm seam that can solve the conflict.
5. Implement only the smallest lifecycle/arbitration change necessary to let Camera record and then restore wake-word listening.

Do not redesign the wake-word detector, change model/runtime/provider policy, or fold battery optimisation into this issue.

## Required behaviour

The target user journey is:

```text
Hey Jandal enabled
  -> open stock Camera
  -> start video recording with audio successfully
  -> Jandal wake capture yields/suspends as required
  -> stop Camera recording
  -> Hey Jandal automatically resumes
  -> wake phrase works again without settings changes
```

If Android provides an explicit recording-silenced/configuration signal on the affected device, prefer using that existing platform signal. If MagicOS rejects Camera before such a signal reaches Jandal, use the smallest reliable application lifecycle/audio-arbitration mechanism demonstrated by device evidence.

## Shared-path regression checks

Because the fix touches wake-word microphone ownership/re-arm behaviour, regression-check only the existing consumers that pass through that seam:

- normal Hey Jandal detection and wake -> STT handoff;
- enable/disable Hey Jandal;
- wake-word service re-arm after a completed command;
- push-to-talk / assistant voice capture taking over the microphone;
- stop/restart behaviour after microphone permission changes where already covered.

Do not broaden this into a general voice architecture audit.

## Acceptance criteria

- [ ] Issue is reproduced on current `main` on Honor Magic 8 Pro, with logs/audio state sufficient to identify the arbitration path.
- [ ] With Hey Jandal enabled, the stock Honor Camera can start and complete a video recording with audio.
- [ ] Jandal does not keep or immediately reclaim the microphone in a way that prevents Camera recording.
- [ ] After Camera recording ends, Hey Jandal automatically returns to a functional listening state without requiring an app restart or settings toggle.
- [ ] A subsequent natural `Hey Jandal` activation succeeds on the Honor Magic 8 Pro.
- [ ] Existing wake -> STT microphone handoff remains functional.
- [ ] Existing manual Jandal voice capture remains functional.
- [ ] Repeated Camera start/stop cycles do not leave the wake service permanently stopped or create microphone contention loops.
- [ ] No wake-word model, thresholds, provider, verifier, or battery optimisation changes are bundled into the fix unless device evidence proves one is strictly required.

## Release impact

**Launch-blocking while Hey Jandal remains release-visible.**

Hey Jandal is opt-in and disabled by default, but once enabled it must not break a normal core phone capability such as stock-camera video recording. Resolve this with a focused fix or consciously change the v0.1 product/release posture for Hey Jandal before release.

lokhor:
Fixes #1502

## Root cause

On the HONOR BKQ-N49 (Magic 8 Pro, MagicOS 10.0.0.199, Android 16), MagicOS refuses to start stock
camera video recording while **any background app holds a `VOICE_RECOGNITION` capture**:

> "A recording app is already running in the background (Recorder, Call). To record a video, close
> the app first."

Hey Jandal's wake detector holds exactly such a capture continuously, so "Hey Jandal" and stock
camera video recording were mutually exclusive. The camera app receives no platform callback that
could warn Jandal in time — audio recording-configuration callbacks, audio focus, audio mode and
routing changes, and `CameraManager` callbacks were all evaluated on device during the
investigation and provide no usable pre-recording signal. Asserting Jandal as the Android
assistant/VoiceInteractionService does not change the OEM policy either (re-tested: the camera was
still blocked 2/2), and the only audio-source alternative that allows coexistence
(`VOICE_PERFORMANCE`) destroys wake detection on this device (0/9 activations vs 5/9 with
`VOICE_RECOGNITION`), so the source is unchanged.

The one mechanism proven to work is to learn that the camera is coming **before** it needs the
microphone, and release Jandal's captures proactively.

## Approved product decision

Use `UsageStatsManager` foreground detection to yield the microphone, scoped to the exact device
class proven by physical evidence — **HONOR + model `BKQ-N49`, camera package
`com.hihonor.camera`**. No generalisation to other Honor models, other MagicOS releases, other
camera apps, or arbitrary microphone users. Usage Access is an accepted requirement for Hey Jandal
on that device; every other device keeps today's behaviour (no special access, no prompt, no
polling, unchanged wake lifecycle).

## Implementation

`core/voice` (new):

- `HonorCameraCoexistence` — device/package predicate, Usage Access (AppOps) readiness, the
  "may this device hold the wake microphone at all" gate, the packaged Usage Access settings intent,
  and `mayRearmAfterCamera(...)` encoding every re-arm condition.
- `CameraForegroundMonitor` + `ForegroundPackageTracker` + `UsageStatsForegroundEventSource` —
  polls foreground-activity transitions at the proven 2 s cadence, reports only transitions of the
  camera in/out of the foreground, isolates failures (a failing query or callback is logged and
  retried on the next cadence, never terminating monitoring), propagates cancellation, and holds
  only the current foreground package name.

`WakeWordService` (new behaviour):

- Refuses to run wake capture at all when Usage Access is missing on the affected device
  (refusal happens after the foreground promotion, because a service started with
  `startForegroundService()` must reach `startForeground()` or Android kills the process with
  `ForegroundServiceDidNotStartInTimeException` — a defect the first physical build exposed).
- Starts exactly one camera monitor when the service is active on the affected device.
- Camera foreground → latch suspension, then release **all** Jandal-owned microphone captures: the
  wake detector *and* any voice capture left over from a wake→STT or push-to-talk session (whose
  capture is actually held by the recognizer process). The service and its ongoing notification
  stay alive; the notification text becomes "Paused while Camera is in use" and returns to
  "Listening for wake word…" when the detector is armed again.
- Camera background → re-arm only when Hey Jandal is still enabled, microphone permission and
  Usage Access remain, no voice session or wake handoff is active, and the camera is gone.
  The wake-session retry loop is closed while suspended, so a yield cannot be undone by attempt 2.
- Usage Access revoked while running → release the microphone and stop; the Voice settings screen
  turns Hey Jandal off and explains what is needed.

`feature:settings` — new Usage Access prompt on the affected device only
(`hey_jandal_usage_access_dialog`), with the specified copy, an `Open Usage Access` action that
opens MagicOS's Usage Access surface for Jandal AI, a `Not now` action, re-check on return, and an
enforcement path that disables an already-enabled Hey Jandal rather than letting the UI claim it is
listening. The AppOps check is cached per resume, not per recomposition.

## Privacy

Only the camera's foreground state is evaluated. The current foreground package name is held in
memory for the poll window, unrelated usage events are discarded immediately, and nothing is
persisted, exported, added to diagnostics/memories/analytics, or transmitted. Production logging
never records unrelated package names. The bounded lookback windows are the 60 s initial-state
window and the 3 s immediate re-arm window.

## Automated tests

1578 unit tests pass (0 failures), lint clean with no issues in the changed files.

- `HonorCameraCoexistenceTest` (19) — device predicate incl. other Honor models and the S23U/S21,
  Usage Access mapping (`MODE_ALLOWED` only), exception paths, and the bounded immediate re-arm
  window.
- `CameraForegroundMonitorTest` (12) — one suspension per foreground transition, idempotent
  repeated events, re-arm on exit, another app taking over counts as an exit, immediate suspension
  when the monitor starts with the camera already foreground, window discipline, query-failure and
  callback-failure recovery, revocation fail-safe, cancellation stops polling, cancellation is not
  swallowed, and tracker semantics.
- `WakeCaptureRearmPolicyTest` (6) — each re-arm condition independently blocks re-arm.
- `HonorCameraWakeReleaseTest` — the suspension latch is set before any capture is released.
- `WakeCommandHandoffTest` — a yielded microphone closes the STT retry loop (verified to fail
  without the guard).

## Physical validation — HONOR BKQ-N49, MagicOS 10.0.0.199, Android 16

Source `218cf29a`, APK `110771c1…e110788`, Usage Access granted, Hey Jandal enabled.
Logcat is not readable on this device (buffers show 0 B readable over adb), so evidence comes from
`dumpsys activity services`, `dumpsys media.audio_policy` capture ownership, the debug target-event
journal, UI dumps and the recorded media itself.

| Check | Result |
|---|---|
| A. First-time setup | Prompt shown with the specified copy and test tags; `Open Usage Access` opened `com.android.settings/.Settings$AppUsageAccessSettingsActivity` for Jandal AI; after granting and returning, Hey Jandal auto-enabled itself (pref true, service running, capture active). An already-enabled Hey Jandal without Usage Access was disabled and the prompt shown — the update path. |
| B. Camera happy path | Stock camera recorded 128 s / 1.1 GB with **no contention dialog**; AAC 48 kHz stereo, mean −41.7 dB, max −17.1 dB (real audio); Jandal re-armed after the camera exited. |
| C. Immediate-record race | 3/3 runs with Record tapped ~1.1 s after the camera activity resumed: no dialog, all finalised (~115 MB), audio verified (mean −44.6 / −42.8 dB). Measured suspend latency 1.8 s after the camera takes the foreground; **no cadence change needed**. |
| D. Repeated cycles | 4/4 cycles: recording started, no dialog, real audio, detector re-armed, single service/detector instance throughout. |
| E. Wake recovery after camera | Control 2/4 vs after-camera 4/4 (confidence 0.54–0.85) with full `ACTIVATION_CANDIDATE → VERIFIED_ACTIVATION → WAKE_CALLBACK_INVOKED → VOICE_SESSION_STARTED → STT_READY → CUE_PLAYBACK_STARTED`; one trial completed the whole command path to `SESSION_COMPLETED → DETECTOR_REARMED → STAGE3_READY`. |
| F. Wake/STT session then camera | With a Jandal voice session active whose capture was held by the recognizer process (`com.google.android.tts`, uid 10190), the camera recorded 68.6 MB with real audio (mean −44.0 dB) and the recognizer's capture was gone during recording — the lingering-recognizer case from the spike. |
| G. Manual PTT then camera | PTT active → wake detector yielded; camera recorded 50.9 MB with real audio (mean −45.6 dB); PTT released; detector re-armed after the camera. |
| H. Disable while suspended | Hey Jandal disabled → camera cycle still recorded (122.5 MB) → service stayed stopped, zero captures, stored preference unchanged, and **no detector journal events at all** (no re-arm). |
| I. Usage Access revocation | Revoked externally while listening: within 4 s the service stopped and the microphone was released, no crash; Voice settings showed the Usage Access prompt and corrected the stored preference. Re-granting via `Open Usage Access` restored wake listening. |
| J. Microphone permission regression | Revoke → service stopped, capture released, no crash. Grant → recovered (service running, capture active). |
| CPU/polling | Isolated monitor-on vs monitor-off over 120 s windows with the camera foreground: **0.249% vs 0.150% of one core → the monitor costs ~0.10% of one core** (one `queryEvents` per 2 s), matching the spike's ~2 ms/query. |

Regression checks: normal listening, wake detection, wake→STT, command completion → re-arm, manual
PTT, enable/disable, microphone revoke/restore, service restart and an already-enabled update were
all exercised above.

## Limitations

- The **sideloaded debug APK** is subject to Android 13+ restricted settings, so MagicOS blocks the
  Usage Access toggle in system settings until "Remove restriction" is confirmed with the
  lock-screen password ("Controlled by restricted setting"). A Play-installed release build is not
  subject to restricted settings. To keep validating the product behaviour, Usage Access was granted
  via the equivalent AppOps grant; the app-side flow (prompt, settings surface, re-check on return,
  enable, enforcement, revocation fail-safe) was exercised unchanged.
- Lock-screen camera is not reachable over ADB on this device (Honor's double-power gesture opens
  Wallet, and launching the camera under a secure keyguard resumes then immediately stops), so the
  workaround was not exercised from the lock screen. The monitor is foreground-package based and is
  not expected to depend on unlock state, but that is unverified.
- The Honor acoustic wake fixture is marginal on this device (~50% detection), so rates above are
  directional; the comparison is control vs after-camera on the same day, not a reliability
  qualification. Wake trials ran with Bluetooth temporarily disabled because the fixture harness
  refuses to play through a Bluetooth output route; Bluetooth was restored afterwards.

## Store/privacy follow-up

Narrow factual comments were added to #1260 (privacy policy / Data Safety / in-app summary) and
#1261 (App content declarations and reviewer access) recording the new special access, its purpose,
the device scope, local-only processing and the reviewer step.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -34,6 +34,16 @@
     <uses-permission android:name="android.permission.CALL_PHONE" />
     <!-- Wake word foreground service requires microphone access type (API 35+) -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
+    <!--
+        #1502: special (AppOps) access, not a runtime permission. Requested only to let Hey Jandal
+        detect the Honor stock Camera entering the foreground on the affected device (HONOR
+        BKQ-N49) so wake capture can release the microphone before the camera records. The user
+        grants it from system settings and can revoke it at any time. Only the Camera's own
+        foreground state is evaluated; no usage history is stored or transmitted.
+    -->
+    <uses-permission
+        android:name="android.permission.PACKAGE_USAGE_STATS"
+        tools:ignore="ProtectedPermissions" />
 
     <!--
         Android 11+ package visibility: declare which implicit intents we resolve
--- a/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
+++ b/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
@@ -648,24 +648,6 @@ class WakeWordService : Service() {
             return START_NOT_STICKY
         }
 
-        // #1502: on the Honor device where the camera contends for the microphone, wake capture may
-        // only run with Usage Access — without it Jandal cannot learn that the camera needs the
-        // microphone and would silently block video recording. Refuse before promoting to the
-        // microphone foreground state so no misleading "listening" notification is ever posted.
-        if (!HonorCameraCoexistence.isWakeCaptureAllowed(this)) {
-            Log.w(
-                TAG,
-                "WakeWordService: Usage Access missing on ${Build.MANUFACTURER} ${Build.MODEL} " +
-                    "— refusing to start wake capture",
-            )
-            stopSelf(startId)
-            AcousticJournalBridge.record(
-                type = AcousticEventType.SERVICE_ERROR,
-                metadata = { mapOf("category" to "usage_access_missing") },
-            )
-            return START_NOT_STICKY
-        }
-
         val foregroundStarted = tryPromoteToMicrophoneForeground(
             promote = { startForeground(NOTIFICATION_ID, buildNotification(NOTIFICATION_TEXT_LISTENING)) },
             onRejected = { error ->
@@ -682,6 +664,28 @@ class WakeWordService : Service() {
         )
         if (!foregroundStarted) {
             stopSelf(startId)
+            return START_NOT_STICKY
+        }
+
+        // #1502: on the Honor device where the camera contends for the microphone, wake capture may
+        // only run with Usage Access — without it Jandal cannot learn that the camera needs the
+        // microphone and would silently block video recording.
+        //
+        // This refusal deliberately runs *after* the foreground promotion: a service started with
+        // startForegroundService() must reach startForeground() or Android kills the process with
+        // ForegroundServiceDidNotStartInTimeException. The service is stopped immediately after, so
+        // the ongoing notification never outlives the refusal.
+        if (!HonorCameraCoexistence.isWakeCaptureAllowed(this)) {
+            Log.w(
+                TAG,
+                "WakeWordService: Usage Access missing on ${Build.MANUFACTURER} ${Build.MODEL} " +
+                    "— refusing to start wake capture",
+            )
+            stopSelf(startId)
+            AcousticJournalBridge.record(
+                type = AcousticEventType.SERVICE_ERROR,
+                metadata = { mapOf("category" to "usage_access_missing") },
+            )
             return START_NOT_STICKY
         }
 
--- a/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
+++ b/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
@@ -883,8 +883,10 @@ class WakeWordService : Service() {
     /** Re-arms [wakeWordDetector] with the standard callbacks. */
     private fun rearmDetector() {
         // #1502: a camera-driven suspension is a latch — no re-arm path may take the microphone
-        // back while the Honor Camera needs it.
-        if (isWakeCaptureSuspended) {
+        // back while the Honor Camera needs it. The latch is set on the monitor's next poll, so a
+        // session ending inside that gap is covered by an immediate, bounded foreground check;
+        // otherwise that session end would hand the microphone straight back to Jandal.
+        if (isWakeCaptureSuspended || HonorCameraCoexistence.shouldWithholdWakeCapture(this)) {
             Log.i(TAG, "WakeWordService: wake capture suspended — not re-arming")
             return
         }
--- a/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
+++ b/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
@@ -20,6 +20,7 @@ import com.kernel.ai.core.voice.NO_SPEECH_WINDOW_EXHAUSTED
 import com.kernel.ai.core.voice.AcousticJournalBridge
 import com.kernel.ai.core.voice.CameraForegroundMonitor
 import com.kernel.ai.core.voice.HonorCameraCoexistence
+import com.kernel.ai.core.voice.UsageStatsCurrentForegroundSource
 import com.kernel.ai.core.voice.UsageStatsForegroundEventSource
 import com.kernel.ai.core.voice.WakeWordPreferences
 import com.kernel.ai.core.voice.containsWakePhrase
@@ -498,13 +499,6 @@ private const val NOTIFICATION_TEXT_CAMERA_PAUSED = "Paused while Camera is in u
 internal const val HONOR_CAMERA_POLL_INTERVAL_MS = 2_000L
 
 /**
- * #1502: bounded lookback used only for the monitor's *initial* foreground state, so a service
- * that starts while the camera is already foreground still releases the microphone. Every later
- * poll reads only the window since the previous poll.
- */
-internal const val HONOR_CAMERA_INITIAL_LOOKBACK_MS = 60_000L
-
-/**
  * #1502: release every Jandal-owned microphone capture so the Honor Camera can record.
  *
  * The suspension latch is set *before* any capture is released: every re-arm path (voice-session
@@ -512,6 +506,10 @@ internal const val HONOR_CAMERA_INITIAL_LOOKBACK_MS = 60_000L
  * take the microphone back while the release is still in flight. On BKQ-N49 the wake detector is
  * not the only Jandal-owned capture that can block the camera — a session left over from a
  * wake→STT flow holds one too, hence [releaseVoiceCapture].
+ *
+ * Both releases are attempted even when one of them throws: a single failing release must not
+ * leave the other capture holding the microphone. The first failure is re-thrown so the caller can
+ * retry the whole release on its next attempt.
  */
 internal fun releaseWakeCaptureForCamera(
     suspendLatch: () -> Unit,
@@ -519,8 +517,42 @@ internal fun releaseWakeCaptureForCamera(
     releaseVoiceCapture: () -> Unit,
 ) {
     suspendLatch()
-    releaseWakeDetector()
-    releaseVoiceCapture()
+    var failure: Exception? = null
+    fun attempt(release: () -> Unit) {
+        try {
+            release()
+        } catch (e: Exception) {
+            Log.w(TAG, "WakeWordService: releasing a microphone capture failed", e)
+            val first = failure
+            if (first == null) failure = e else first.addSuppressed(e)
+        }
+    }
+    attempt(releaseWakeDetector)
+    attempt(releaseVoiceCapture)
+    failure?.let { throw it }
+}
+
+/**
+ * #1502: release Jandal's captures and stop the wake service, even when a release throws.
+ *
+ * Losing the camera-coexistence capability means Jandal can no longer yield the microphone, so
+ * listening must not continue — and a service that keeps running after that loss is a worse
+ * outcome than a release that has to be retried. The release failure is re-thrown after the stop
+ * so callers that retry cleanup (the camera monitor) still see it.
+ */
+internal fun releaseWakeCaptureAndStop(
+    releaseCaptures: () -> Unit,
+    stopService: () -> Unit,
+) {
+    var failure: Exception? = null
+    try {
+        releaseCaptures()
+    } catch (e: Exception) {
+        Log.w(TAG, "WakeWordService: releasing microphone captures failed", e)
+        failure = e
+    }
+    stopService()
+    failure?.let { throw it }
 }
 
 /**
@@ -707,7 +739,7 @@ class WakeWordService : Service() {
         }
 
         Log.i(TAG, "WakeWordService: starting wake word detection")
-        rearmDetector()
+        startWakeCapture()
 
         // Automatically yield the AudioRecord whenever another voice session is active.
         eventCollectorJob = serviceScope.launch {
@@ -753,6 +785,23 @@ class WakeWordService : Service() {
     // ── Honor camera coexistence (#1502) ───────────────────────────────────────
 
     /**
+     * #1502: take the microphone, but only once wake capture may legitimately run.
+     *
+     * Unaffected devices arm straight away. On the affected Honor device the camera monitor owns
+     * the first arm: it establishes the current foreground package from usage aggregation and
+     * reports it, so a service or process restart while the Honor Camera has already been in front
+     * cannot take the microphone first and fix it on a later poll — a camera foreground for longer
+     * than any event window appears in no window at all.
+     */
+    private fun startWakeCapture() {
+        if (HonorCameraCoexistence.isAffectedDevice(this)) {
+            startHonorCameraMonitor()
+        } else {
+            rearmDetector()
+        }
+    }
+
+    /**
      * #1502: start the single Honor-camera foreground monitor.
      *
      * Only runs on the proven affected device (Usage Access is already verified by the refusal
@@ -767,7 +816,7 @@ class WakeWordService : Service() {
                 source = UsageStatsForegroundEventSource(applicationContext),
                 targetPackage = HonorCameraCoexistence.CAMERA_PACKAGE,
                 pollIntervalMs = HONOR_CAMERA_POLL_INTERVAL_MS,
-                initialLookbackMillis = HONOR_CAMERA_INITIAL_LOOKBACK_MS,
+                currentForegroundPackage = UsageStatsCurrentForegroundSource(applicationContext),
                 isReady = { HonorCameraCoexistence.hasUsageAccess(this@WakeWordService) },
                 onCameraEntered = { suspendForHonorCamera() },
                 onCameraExited = { resumeAfterHonorCamera() },
@@ -822,7 +871,7 @@ class WakeWordService : Service() {
             Log.i(TAG, "WakeWordService: camera left foreground but wake capture may not re-arm yet")
             return
         }
-        Log.i(TAG, "WakeWordService: camera left foreground — re-arming wake capture")
+        Log.i(TAG, "WakeWordService: camera not foreground — arming wake capture")
         rearmDetector()
     }
 
@@ -838,12 +887,16 @@ class WakeWordService : Service() {
             type = AcousticEventType.SERVICE_ERROR,
             metadata = { mapOf("category" to "usage_access_revoked") },
         )
-        releaseWakeCaptureForCamera(
-            suspendLatch = { isWakeCaptureSuspended = true },
-            releaseWakeDetector = { wakeWordDetector.stop() },
-            releaseVoiceCapture = { voiceInputController.stopListening() },
+        releaseWakeCaptureAndStop(
+            releaseCaptures = {
+                releaseWakeCaptureForCamera(
+                    suspendLatch = { isWakeCaptureSuspended = true },
+                    releaseWakeDetector = { wakeWordDetector.stop() },
+                    releaseVoiceCapture = { voiceInputController.stopListening() },
+                )
+            },
+            stopService = { stopSelf() },
         )
-        stopSelf()
     }
 
     // ── Detection handoff ──────────────────────────────────────────────────────
@@ -883,9 +936,10 @@ class WakeWordService : Service() {
     /** Re-arms [wakeWordDetector] with the standard callbacks. */
     private fun rearmDetector() {
         // #1502: a camera-driven suspension is a latch — no re-arm path may take the microphone
-        // back while the Honor Camera needs it. The latch is set on the monitor's next poll, so a
-        // session ending inside that gap is covered by an immediate, bounded foreground check;
-        // otherwise that session end would hand the microphone straight back to Jandal.
+        // back while the Honor Camera needs it. On top of the latch, ask whether the camera owns
+        // the foreground right now: that covers a session ending between the camera appearing and
+        // the monitor's next poll, and a restart that lost the latch while the camera was already
+        // in front (see HonorCameraCoexistence.shouldWithholdWakeCapture).
         if (isWakeCaptureSuspended || HonorCameraCoexistence.shouldWithholdWakeCapture(this)) {
             Log.i(TAG, "WakeWordService: wake capture suspended — not re-arming")
             return
--- a/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
+++ b/app/src/main/java/com/kernel/ai/assistant/WakeWordService.kt
@@ -7,6 +7,7 @@ import android.app.PendingIntent
 import android.app.Service
 import android.content.Context
 import android.content.Intent
+import android.os.Build
 import android.os.IBinder
 import android.os.Handler
 import android.os.Looper
@@ -17,7 +18,12 @@ import com.kernel.ai.MainActivity
 import com.kernel.ai.core.voice.AcousticEventType
 import com.kernel.ai.core.voice.NO_SPEECH_WINDOW_EXHAUSTED
 import com.kernel.ai.core.voice.AcousticJournalBridge
+import com.kernel.ai.core.voice.CameraForegroundMonitor
+import com.kernel.ai.core.voice.HonorCameraCoexistence
+import com.kernel.ai.core.voice.UsageStatsForegroundEventSource
+import com.kernel.ai.core.voice.WakeWordPreferences
 import com.kernel.ai.core.voice.containsWakePhrase
+import com.kernel.ai.core.voice.mayRearmAfterCamera
 import com.kernel.ai.core.voice.StartListeningCuePlayer
 import com.kernel.ai.core.voice.StartListeningCueResult
 import com.kernel.ai.core.voice.StartListeningCueContext
@@ -46,6 +52,7 @@ import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.coroutineScope
 import kotlin.coroutines.coroutineContext
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import java.lang.ref.WeakReference
 import java.security.MessageDigest
@@ -393,7 +400,8 @@ internal data class WakeCommandHandoffResult(
  *    keeps the session open across the cue-to-command handoff so the runner's
  *    command is captured in attempt 1 and no retry is needed on the normal path.
  *
- * The bounded retry still runs for genuine post-readiness recognition failures.
+ * The bounded retry still runs for genuine post-readiness recognition failures, but never when
+ * [capturePermitted] reports that the microphone was taken away in the meantime (#1502).
  * On cancellation the active recognizer is stopped so no microphone owner or
  * recognizer is left behind.  [rearmDetector] is invoked exactly once, after the
  * session is terminal (including on cancellation, matching the previous service
@@ -409,6 +417,7 @@ internal suspend fun runWakeCommandHandoff(
     routeTranscript: (String) -> Boolean,
     onSessionTerminal: () -> Unit,
     journal: WakeSessionJournal? = null,
+    capturePermitted: () -> Boolean = { true },
 ): WakeCommandHandoffResult {
     // #1433: a confirmed wake activation must not request live STT capture until the
     // wake detector has completed microphone-resource release.
@@ -421,6 +430,15 @@ internal suspend fun runWakeCommandHandoff(
 
     try {
         val sessionResult = runWakeCaptureSession { attempt ->
+            // #1502: the microphone can be yielded mid-session (Honor Camera taking the
+            // foreground). Starting another attempt would take it straight back and block
+            // camera recording again, so the retry loop is closed instead.
+            if (!capturePermitted()) {
+                throw WakeAttemptCollectionException(
+                    category = "wake_capture_suspended",
+                    cause = IllegalStateException("wake capture suspended before attempt $attempt"),
+                )
+            }
             runWakeAttempt(
                 voiceInputController = voiceInputController,
                 journal = sessionJournal,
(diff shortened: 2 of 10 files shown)


NickMonrad (changes requested):
I found two merge-blocking lifecycle defects in the #1502 implementation.

1. **P1 — Wake capture can re-arm after a sticky service/process restart while Honor Camera is already foreground.**
   - **Location:** `WakeWordService.onStartCommand()` / `rearmDetector()` plus `HONOR_CAMERA_INITIAL_LOOKBACK_MS`, and `HonorCameraCoexistence.shouldWithholdWakeCapture()`.
   - **Failure scenario:** the service is `START_STICKY`. If Jandal is restarted while `com.hihonor.camera` has already been foreground for longer than the 60 s initial lookback, the process-local `isWakeCaptureSuspended` latch has been reset. `onStartCommand()` calls `rearmDetector()` before `startHonorCameraMonitor()`, and the immediate guard only looks back 3 s. The monitor then reconstructs state from only 60 s of usage events, sees no old Camera-enter event, and never suspends. Jandal therefore holds `VOICE_RECOGNITION` again while Camera is still open, so the next video recording can reproduce the exact #1502 failure. The test `a camera launch older than the window is left to the suspension latch` assumes a latch that does not survive process/service restart.
   - **Why this violates the issue:** #1502 requires Camera coexistence and the implementation prompt explicitly included service restart as a shared-path regression. The current startup path treats “no recent enter event” as “Camera not foreground,” which is not equivalent after process state is lost.
   - **Smallest remediation:** on the affected device, establish a trustworthy *current* Camera foreground state before arming the detector on service startup/restart; do not rely on a finite recent-event window plus an in-memory latch. Also avoid calling `rearmDetector()` until that initial state is known. Add a regression where the service restarts after the Camera has been foreground longer than the current initial lookback and assert no wake capture starts until Camera exits.

2. **P1 — Usage Access revocation can permanently lose the fail-safe if the release callback throws.**
   - **Location:** `CameraForegroundMonitor.run()` in the `!isReady()` branch (`guard { onUnavailable() }; return`) and `WakeWordService.onHonorCameraCoexistenceUnavailable()`.
   - **Failure scenario:** Usage Access is revoked while Jandal owns a capture. `onHonorCameraCoexistenceUnavailable()` must stop the detector, stop any voice capture, then `stopSelf()`. If either release operation throws, `guard()` swallows the exception and `run()` returns anyway. The monitor is now gone, `stopSelf()` may never have executed, and any capture after the failing operation may remain alive with no future poll to retry the fail-safe. This is exactly the class of lifecycle-exception problem the spike required production code to harden against.
   - **Why this violates the issue:** the approved design requires revocation to fail safe and specifically requires pause/resume/release handling to be exception-safe; one failed lifecycle operation must not silently terminate monitoring while the microphone can still be owned.
   - **Smallest remediation:** only terminate the monitor after the unavailable callback completes successfully, or retry the unavailable callback on the next cadence when it fails. In the service callback, independently attempt both capture releases and ensure `stopSelf()` still executes even if one release throws. Add a focused test where `onUnavailable` fails once and then succeeds, proving the monitor does not disappear after the first failure.

Everything else I reviewed is consistent with the approved scope: exact `HONOR` + `BKQ-N49` gating, no polling on unaffected devices, no wake-source/threshold changes, the Usage Access UX is device-scoped, CI/evidence generation are green, and the physical Honor results are strong. Fix the two lifecycle holes above and this should be ready for re-review.

lokhor:
## Remediation of review 5175449011 — code complete, physical re-validation still open

**Head:** `1f84611e` (on top of `218cf29a`)

### Finding 1 — restart can lose the Camera's protection

* The current foreground package is now resolved from usage **aggregation** (most recently visible package over a bounded day range) instead of a finite recent-transition window, and that resolution is shared by the re-arm guard and the monitor's startup state. A camera that has been in front for longer than any window has no recent enter event, so this is the only way to see it.
* On the affected device the **monitor owns the first arm**: `onStartCommand` no longer arms, the monitor resolves the current state and reports it (forced initial `reconcile`), and `resumeAfterHonorCamera` is what arms. Capture is therefore never taken first and corrected on a later poll.
* The re-arm guard asks for the current state rather than "did the camera enter recently"; an unreadable or failed state **withholds** capture (the two costs are asymmetric: withholding quiets the wake word until the next transition, arming blocks the camera).
* `HONOR_CAMERA_INITIAL_LOOKBACK_MS` / `CAMERA_JUST_ENTERED_LOOKBACK_MS` and the invalid "an older camera launch is covered by the suspension latch" assumption are gone.
* Empirical basis on BKQ-N49 (via `dumpsys usagestats`): a camera in front for over 100 s is the most recently visible package, and the package that takes over from it replaces it within 1.5 s.

### Finding 2 — revocation fail-safe could disappear after one exception

* `CameraForegroundMonitor` terminates only after the unavailable callback **succeeds**; a failed cleanup is retried on the next cadence, and cancellation still propagates (the swallowing `guard` is gone).
* `releaseWakeCaptureForCamera` now attempts **both** releases even when one throws, and re-throws the first failure so the caller can retry the whole release.
* New `releaseWakeCaptureAndStop` guarantees `stopSelf()` runs even when a release fails, then re-throws so the monitor can retry.

### Automated evidence

* `:core:voice:testDebugUnitTest` and `:app:testDebugUnitTest` pass; full `./gradlew test` passes; `:app:lintDebug` clean.
* New/updated coverage: restart with a camera in front before the monitor started (nothing arms while it is in front; the exit is what arms), clean startup reports the state so the caller may arm, unreadable startup state retried without consulting a stale window, cleanup retried until it succeeds (and stops monitoring only then), cancellation during cleanup not swallowed, both releases attempted when one throws, `stopSelf` guaranteed when a release throws, current-state guard semantics (camera / other app / unknown / throwing), most-recently-visible resolution.
* **Both new groups were verified to fail against the pre-fix behaviour** (seeding neutered, and cleanup made non-retryable) before being restored.

### Physical status — incomplete, one unexplained observation

Verified on BKQ-N49 with this head installed:
* camera foreground (after a fresh process start) → wake capture released and the service stays up;
* service alive across a long camera session (>100 s).

**Not yet verified**, and the reason this is not ready for re-review:
* the restart path end-to-end — `am kill` cannot kill the app because the wake service is a foreground service, so the process restart has to be driven another way (a sticky restart or a service re-creation);
* one run ended with the service still reporting "Paused while Camera is in use" after the camera had left, i.e. **no re-arm**. That run also had known defects of its own (the device was locked for the first attempt, so the camera resumed behind the keyguard without a usable foreground state, and `am kill` was a no-op), so it is not yet attributable. The same run's automated equivalents pass.

The decisive open question is whether the **app-side** `queryUsageStats` aggregation is live on this device. My freshness measurements came from `dumpsys usagestats`, which prints in-memory stats that the public API may or may not merge; if the API is only flushed periodically, this approach cannot reconstruct the current package after an arbitrarily long camera session, and per the review's instruction I will **report that limitation rather than add persistence or broader heuristics**. The plan is an on-device probe using the app's own UID (instrumentation test) to settle it.

Physical re-validation resumes when the test device is available again.

Do not merge — this still needs the physical restart check and re-review.

lokhor:
## Physical validation of the review's Finding 1 — one approach replaced

**Head:** `f02d920e` (on top of `1f84611e`)

### What the physical run showed for the previous head

The previous head resolved the current foreground package from usage **aggregation** and put that resolution behind both the monitor's startup state and the re-arm guard. On the affected device (BKQ-N49) that resolution does not work, and the review's P1 reproduces:

| step | expected | observed on `1f84611e` |
|---|---|---|
| Camera takes the foreground | mic released | released ✓ (`Paused while Camera is in use`) |
| Jandal's process dies while the Camera is in front | restarted service must not take the mic | **mic taken** ✗ — a new `AudioRecord` for Jandal's uid with the camera in front |
| Camera leaves (HOME) | re-armed | already held ✗ (nothing to re-arm) |
| Camera returns | released | released ✓ |

A separate earlier observation matches the same cause in the other direction: after the camera left, the guard kept withholding and the wake word stayed paused (`Paused while Camera is in use` outlived the camera) — the aggregation still reported the camera.

**Measured cause.** `queryUsageStats` aggregation lags the device. With the camera foreground it reported a package that had not been visible for minutes, and after the camera left it kept reporting the camera. `queryEvents` is live on the same device (the monitor's 2 s poll sees the camera's enter and exit within a sub-cadence). Earlier evidence for the aggregation came from `dumpsys usagestats`, which prints an in-memory view the public API does not expose — that is my error, and it is why the previous attempt looked sound before it was exercised physically.

### The change

* `UsageStatsCurrentForegroundSource` now replays foreground **events** instead of reading aggregation, sharing the monitor's own transition rule (`applyForegroundTransitions`) so there is only one definition of "who is in front".
* The startup state replays a **6 h** lookback (`SEED_LOOKBACK_MS`) — it has to outlast a camera session that is already running, not the poll cadence. That is what makes the restart case work.
* The re-arm guard replays **5 s** (`REARM_GUARD_LOOKBACK_MS`) — one poll cadence plus event-delivery latency — for the paths the monitor does not drive (a voice session ending, the wake handoff finishing).
* A window with no foreground activity is now a **valid** answer meaning "the camera produced no event in this window", so it clears the guard and lets a startup arm. Only a query that cannot run at all (no usage-stats service, unreadable result) withholds capture. Previously "no package" withheld, which is what left the wake word paused.
* `UsageStatsForegroundEventSource` now throws instead of returning an empty list when the usage-stats service or the query is unavailable, so "unreadable" can never be mistaken for "nothing in front".

### Physical evidence on `f02d920e` (two independent runs)

Both runs, identical: service foreground and alive throughout, acceptance read from the wake notification text and from whether Jandal's uid holds an active `AudioRecord` (app logs are Honor-encrypted on this device).

| step | expected | run 1 | run 2 |
|---|---|---|---|
| armed, no camera | mic held | held ✓ | held ✓ |
| Camera takes the foreground | release | released ✓, `Paused while Camera is in use` | released ✓, same |
| **process killed while the Camera is in front** | **must not take the mic** | **no mic taken ✓, still `Paused…`** | **no mic taken ✓, still `Paused…`** |
| Camera leaves (HOME) | re-arm | re-armed ✓, `Listening for wake word` | re-armed ✓, same |
| Camera returns | release again | released ✓ | released ✓ |
| cleanup (camera closed, HOME) | re-arm | re-armed ✓ | re-armed ✓ |

The camera was in the foreground for 40–60 s before each kill, and the kill is `kill -9` from the app's own uid because `am kill` cannot kill a foreground service.

### Residual limitation

A camera session that started earlier than the 6 h lookback **and produced no event since** cannot be reconstructed from any public API — `queryEvents` is window-bounded and aggregation is the only long-range source, which is now measured as unusable here. Real sessions are minutes; 6 h is deliberate headroom. The bound is documented at `SEED_LOOKBACK_MS`.

### Automated

* full `./gradlew test` passes; `:app:lintDebug` clean.
* The two changed expectations fail against the previous behaviour: "no foreground activity in the window does not withhold" and "a startup whose lookback holds no foreground activity arms" both encode the new semantics, and a new test pins the seed window (`NOW - SEED_LOOKBACK_MS .. NOW`) and that a camera which resumed earlier and never left resolves as the current app.
* Finding 2's remediation from `1f84611e` is unchanged; the pass to compile, test, install and physically re-validate was re-run on this head.

Do not merge — needs re-review.

NickMonrad (changes requested):
Focused re-review of only the remediation changes since `218cf29a`.

**Finding 1 is resolved.** The affected-device startup no longer arms first; the camera monitor establishes state before the first arm, and the replacement event-based implementation is backed by two physical BKQ-N49 process-kill runs where Camera remained foreground, Jandal did not reacquire the microphone, and Camera exit re-armed normally. The documented >6 h reconstruction limitation is not a credible v0.1 blocker.

One merge-blocking defect remains from Finding 2:

**P1 — the production revocation callback still defeats the monitor's retry-on-failure contract.**

- **Location:** `WakeWordService.onHonorCameraCoexistenceUnavailable()` / `releaseWakeCaptureAndStop()` together with `CameraForegroundMonitor.releaseAndStop()`.
- **Concrete failure:** the monitor correctly catches a thrown `onUnavailable()` and plans to retry on the next 2 s cadence. But the production callback calls `releaseWakeCaptureAndStop()`, which executes `stopSelf()` even when `releaseWakeCaptureForCamera()` failed, then rethrows. `stopSelf()` causes `WakeWordService.onDestroy()`, which cancels `serviceScope`; that is the same scope hosting `cameraMonitorJob`. The coroutine therefore cannot survive to the next cadence to perform the retry its unit test assumes. If the failed operation was `voiceInputController.stopListening()`, the recognizer capture can remain outside `WakeWordService` while the only retry mechanism is cancelled.
- **Why this still violates #1502:** the original finding required a transient release failure not to permanently remove the fail-safe while a Jandal-owned capture can remain. The monitor unit test proves retry with a synthetic callback that does not stop its owning scope, but the real service callback has different lifecycle semantics.
- **Smallest remediation:** make the production lifecycle consistent with the retry contract. Do not terminate/cancel the monitor's owning service before failed capture cleanup has been retried successfully. Once cleanup succeeds, stop the service. Preserve the requirement that both capture releases are attempted on every cleanup attempt and cancellation still propagates. Add a focused service/monitor integration test (or equivalent extracted seam) proving: first production cleanup attempt fails, the monitor remains alive, the next cadence retries, second cleanup succeeds, and only then is service stop requested.

CI for current head `f02d920e` is still running at review time, but this defect is independent of CI outcome. No other material remediation defect found.

lokhor:
## Both review findings — remediation and evidence (head `bb75e9e0`)

## Finding 1 — restart can lose Camera foreground protection

**How the current Camera state is established before arming.** On the affected device nothing arms from `onStartCommand`: `startWakeCapture()` starts the `CameraForegroundMonitor` instead, and the monitor's first action is to resolve the current foreground package (`seed()`), report it, and only then call back. `resumeAfterHonorCamera()` — the only path that arms on this device — is reached from the monitor's exit transition, so capture is never taken first and corrected on a later poll. The resolution replays foreground **events** over a 6 h window (`SEED_LOOKBACK_MS`) through the same `applyForegroundTransitions` rule the tracker uses, so there is one definition of "who is in front". No history is persisted: no database, no preference, nothing written at all.

**Why the >60 s Camera case is now safe.** The startup window is hours, not the poll cadence, so a Camera that resumed before the service existed is still inside it; `queryEvents` is live on this device (measured), so the ENTER is present; and no event in the window is a valid "the Camera is not in front" answer that lets a clean startup arm normally. An unreadable query withholds capture instead of arming.

**Tests** (`core/voice/src/test/.../CameraForegroundMonitorTest.kt`):

| test | defends |
|---|---|
| `a camera in front before the monitor started keeps wake capture suspended until it leaves` | nothing arms before the startup state is known; the Camera's exit is what arms |
| `the current package is replayed from the events that are still readable` | the window reaches back at least an hour, and a Camera that resumed earlier and never left resolves as the current package |
| `a startup whose lookback holds no foreground activity arms` | an empty window is a valid state, not a permanent withhold |

Mechanical check that these fail against the pre-fix behaviour: shrinking the lookback to the old poll-sized window (`SEED_LOOKBACK_MS = 2_000L`) fails the class; removing `tracker.seed(current)` fails the class. Both mutations were applied and reverted.

**Physical (head `f02d920e`, production code identical to `bb75e9e0`).** Two independent runs: Camera foreground → mic released; `kill -9` of Jandal's process with the Camera still in front → the restarted service **held no microphone** and stayed `Paused while Camera is in use`; Camera exit (HOME) → re-armed; Camera return → released again. The previous head took the microphone in the same step. The Camera had been in front ~16–20 s before each kill in these runs; a longer-hold (5 min) variant is still to be run — the device left ADB range.

## Finding 2 — Usage Access revocation fail-safe can disappear

* `CameraForegroundMonitor.releaseAndStop()` returns true only once `onUnavailable()` has actually completed; a throwing cleanup keeps the monitor in the unavailable state and retries it on the next cadence. Cancellation still propagates.
* `releaseWakeCaptureForCamera` sets the suspension latch first, attempts **both** releases, logs and aggregates failures (`addSuppressed`), and re-throws the first failure so the caller retries the whole release.
* `releaseWakeCaptureAndStop` always calls `stopSelf()` even when a release throws, then re-throws so the monitor retries.

**Tests:** `a failed unavailable cleanup is retried instead of dropping the fail-safe` (first attempt throws → not terminated → retried → second success terminates), `cancellation is not swallowed while cleaning up after lost usage access`, `suspension is latched before the wake detector and voice captures are released`, `a failed wake detector release still releases the voice capture and reports the failure`, `a failed wake detector release still stops the service`, `a successful release stops the service without reporting a failure`.

Mechanical check: making `releaseAndStop()` return true on a failed cleanup fails the monitor class; reverting `releaseWakeCaptureForCamera` to unguarded sequential releases fails `HonorCameraWakeReleaseTest`. Both mutations were applied and reverted.

## Unchanged

Usage Access UX, device predicate, manifest, wake model/audio source, thresholds, polling architecture, privacy/store handling.

lokhor:
### Finding 1 — focused Honor restart validation (Camera held 300 s)

Device BKQ-N49 / Honor, head `f02d920e` (production code identical to `bb75e9e0`). The brief's sequence, with the Camera held **300 s** so the session comfortably exceeds any window the previous implementation reconstructed from, and a `kill -9` while the Camera owns the screen:

| phase | foreground | Jandal uid holds an active `AudioRecord` |
|---|---|---|
| armed baseline | launcher | **yes** — `Listening for wake word` |
| Camera foreground, 30 s → 300 s (10 samples) | Honor Camera | **no** at every sample — `Paused while Camera is in use`; from 90 s on the Camera's own uid (10247) holds the input |
| `kill -9` of Jandal's process, Camera still front, +5 → +30 s (6 samples) | Honor Camera | **no** at every sample — `Paused while Camera is in use` |
| Camera exits (HOME), +5 → +20 s | launcher | **yes** — `Listening for wake word` |

The reviewed head took the microphone back in the kill row; this head does not, and re-arms only once the Camera has left. Setting the screen timeout for the hold was my omission on the first attempt of this run — corrected before the recorded run, so the Camera genuinely held the foreground for the full 300 s.

**Restore state after the run:** Hey Jandal disabled again, screen timeout 30 s, stay-awake off, Camera closed, device dumps removed.

NickMonrad (changes requested):
Focused re-review of the remediation since the previous review. Finding 1 remains resolved, including the 300 s physical restart validation. One previously identified merge-blocking defect is still unresolved:

**P1 — Usage Access revocation cleanup still cancels its own retry path.**

- **Location:** `WakeWordService.onHonorCameraCoexistenceUnavailable()` / `releaseWakeCaptureAndStop()`, together with `CameraForegroundMonitor.releaseAndStop()` and `WakeWordService.onDestroy()`.
- **Concrete failure:** `CameraForegroundMonitor` is designed to retry `onUnavailable()` on the next poll when cleanup throws. But the production callback still calls `releaseWakeCaptureAndStop()`, which calls `stopSelf()` even when `releaseWakeCaptureForCamera()` failed, then rethrows. `stopSelf()` tears down `WakeWordService`; `onDestroy()` cancels `serviceScope`, which owns `cameraMonitorJob`. The monitor therefore cannot survive to perform the retry. If `voiceInputController.stopListening()` is the operation that failed, the recognizer-owned capture can remain while the only retry mechanism has been cancelled.
- **Why this violates #1502:** the prior finding required transient cleanup failures to preserve the fail-safe until all Jandal-owned microphone captures are actually released. The current unit test exercises a synthetic callback that does not stop the monitor's owning scope, so it does not cover production lifecycle semantics.
- **Smallest remediation:** on Usage Access loss, attempt both capture releases but only call `stopSelf()` after cleanup succeeds. If cleanup throws, leave the service/monitor alive so the next cadence retries. Add one focused production-lifecycle seam/integration test proving first cleanup fails -> service is not stopped -> next cadence retries -> cleanup succeeds -> service stop is then requested.

The latest commit after `f02d920e` only strengthens the startup-lookback test and does not alter this production cleanup path. CI is green on current head `bb75e9e0`. No new scope or additional findings.

lokhor:
## Usage Access cleanup lifecycle — remediated (head `1ad8c134`)

**Previous failure path.** `releaseWakeCaptureAndStop()` wrapped the release in a `try`/`catch` and called `stopSelf()` unconditionally, re-throwing afterwards. `stopSelf()` → `onDestroy()` → `serviceScope.cancel()` → `cameraMonitorJob` cancelled, so the monitor never reached its next cadence. A release that threw — most importantly `voiceInputController.stopListening()`, where the recognizer-held capture can outlive the service — was therefore never retried, and the retry contract the monitor documents was unreachable in production. The monitor unit test only retried with a synthetic callback that did not stop its owning service, so it did not model this lifecycle.

**Lifecycle change.** `releaseWakeCaptureAndStop()` is replaced by `releaseWakeCaptureThenStop(suspendLatch, releaseWakeDetector, releaseVoiceCapture, stopService)`, which performs the release first and requests the stop only when it returned:

```
Usage Access lost
  -> latch suspension
  -> attempt wake detector release
  -> attempt voice capture release
  -> release failed?  propagate, DO NOT stop the service, monitor retries next cadence
  -> release completed? stopSelf(), monitoring ends
```

Both releases are still attempted on every attempt (`releaseWakeCaptureForCamera` aggregates failures with `addSuppressed` and re-throws the first), the latch still goes first, and cancellation still propagates — those paths are unchanged. No new scope, manager, worker, retry framework or persistence was added: the monitor's existing cadence is the only retry mechanism. `onHonorCameraCoexistenceUnavailable()` now calls the helper with flat named arguments.

**Regression proving `fail -> remain alive -> retry -> succeed -> stop`.** `HonorCameraWakeReleaseTest.a failed cleanup leaves the service running so the monitor retries it to completion` wires the **real** `CameraForegroundMonitor` to the **real** service-side cleanup, with the monitor running in a scope that `stopService` cancels — production-equivalent, not a standalone helper:

| step | assertion |
|---|---|
| Usage Access revoked | `isReady = false` |
| attempt 1 | suspension latched; detector **and** voice release both attempted; `stops == 0`; `serviceScope.isActive`; monitor job not cancelled |
| next cadence | latch + both releases attempted again |
| attempt 2 succeeds | `stops == 1`, `serviceScope` cancelled, monitor stops |
| later cadences | no further attempts — monitoring terminated normally |

Mutation check: restoring the old unconditional-stop body fails this test and `a failed release does not stop the service`; with the fix both pass. `a failed wake detector release still stops the service` was removed — that expectation *was* the defect. The existing tests for latch-before-release, both-releases-attempted, retry-on-failure and cancellation-not-swallowed are untouched and still pass.

**Validation.** `:core:voice:testDebugUnitTest` ✅ `:app:testDebugUnitTest` ✅ `./gradlew test` ✅ `:app:lintDebug` ✅. Diff for this remediation is two files (`WakeWordService.kt`, `HonorCameraWakeReleaseTest.kt`); no unrelated production changes. Physical re-validation was not required for this change and was not repeated.

NickMonrad (approved):
Focused re-review of the remaining #1502 revocation-cleanup P1 only.

The finding is resolved.

`releaseWakeCaptureThenStop()` now performs the full capture release first and reaches `stopSelf()` only after that release completes successfully. A release failure therefore propagates back through `CameraForegroundMonitor.releaseAndStop()` without cancelling the service scope, so the monitor remains alive and retries on the next cadence. `releaseWakeCaptureForCamera()` still latches suspension first and attempts both capture releases on every attempt.

The new lifecycle regression test covers the production-equivalent sequence that was previously missing: first cleanup fails -> both releases were attempted -> service/monitor remain alive -> next cadence retries -> cleanup succeeds -> only then is service stop requested and the monitor scope cancelled. This directly addresses the prior review finding rather than testing a synthetic callback with different lifecycle semantics.

No new material defect found in the remediation. CI and Docs Drift Check are green on head `1ad8c134`.

Ready to merge.

Issue #1494, opened 2026-08-27
Lists: prove local paired-device sync and conflict convergence


NickMonrad:
Parent: #1490
Depends on: #1491, #1492, #928, #1493
Related: #1489

## Purpose

Prove Jandal's shared-list client/data model end-to-end between **two local Android devices** before any production internet transport is considered.

Use the smallest local/nearby transfer mechanism practical for a prototype. The purpose is to validate change exchange, reconciliation, hierarchy, restore, conflict handling and user-visible convergence — not to create a general networking layer.

### Baseline bidirectional sync

With the same explicitly shared list present on Device A and Device B:

- A adds an item and B receives it;
- B adds an item and A receives it;
- check/uncheck propagates both directions;
- edit/rename propagates where current Lists supports it;
- delete/tombstone propagates;
- explicit restore of a tombstoned item converges using the same stable identity;
- reorder converges according to #1491;
- #928 parent/child relationships and sibling order converge without flattening.

### Hierarchy convergence

Prove at minimum:

1. both devices start with the same flat list;
2. Device A parents one item under another;
3. Device B receives the same two-level hierarchy and order;
4. Device B reorders a child within the sibling group;
5. both devices converge;
6. Device A unparents/promotes the child and both devices converge again.

Also verify that deleting a parent preserves its children as top-level items according to #928 and that the resulting unparent/order changes converge on both devices.

### Offline/reconnect

Prove at minimum:

1. both devices start from the same shared list state;
2. transport is unavailable;
3. A adds `Bread`;
4. B independently adds `Eggs`;
5. transport resumes;
6. both devices converge to a list containing both independent additions without duplicates or loss.

Also cover:

- a same-field item conflict;
- independent edits to different fields on the same item;
- delete versus stale edit;
- explicit restore after delete;
- at least one hierarchy/order change made while a device is offline.

Use the deterministic field/version rules from #1491.

### Duplicate delivery

Deliver the same change/package more than once and prove the receiving store remains correct and unchanged after the first application.

### Local-first behaviour

While no peer/transport is available:

- list create/edit/check/delete/restore/reorder remains usable locally;
- #928 hierarchy operations remain usable locally;
- operations record through the normal #1492 persistence/change path;
- reconnect sends/applies outstanding changes later;
- no list UI is blocked waiting for the other device.

## Integration requirement

Remote changes must be applied through the same authoritative local list persistence seam as local changes.

After applying a remote change:

[...]

Issue #1492, opened 2026-08-27
Lists: migrate persistence to sync-ready stable IDs and change log


NickMonrad:
Parent: #1490
Depends on: #1491
Precedes hierarchy implementation: #928
Related: #897, #917

## Purpose

Implement the minimum Room/data-layer changes required to make existing Jandal Lists safe for later cross-device synchronisation while preserving current local behaviour.

This issue is persistence/data-model work only. It does not implement hierarchy UX, network transport, background sync, provider accounts, hosted services, billing or deployment.

## Required changes

Using the authoritative contract produced by #1491:

- add globally stable list/collection identity separate from editable title and any local Room row identity;
- add globally stable item identity;
- add persisted actor identity/counter state required for deterministic logical version stamps;
- persist the minimum per-field version metadata required by #1491 rather than using wall-clock `updatedAt` as conflict authority;
- represent synchronisable deletion with tombstones and support explicit restore of the same stable identity;
- add the sync-safe sibling order representation required by #1491;
- add nullable stable parent-item identity required for the two-level hierarchy in #928, but do not implement hierarchy UX here;
- persist the canonical shared title separately from any local-only display alias required to resolve name collisions without renaming the shared collection;
- add a bounded local change log/outbox sufficient for later package/local transport;
- migrate existing lists/items without data loss;
- preserve existing read/query behaviour where possible rather than adding a parallel list store.

## Authoritative mutation seam

Current sync-relevant writes originate from multiple paths, including:

- `ListsViewModel`;
- `NativeIntentHandler`;
- `MealPlanSessionRepository`.

#1492 must provide the **smallest shared Room transaction seam** needed so that every sync-relevant state mutation and its corresponding change/outbox record commit atomically.

UI-, voice- and meal-plan-created list changes must not be able to update list state without recording the same logical change for later sync.

Do not move read-only DAO consumers behind a new repository merely for architectural symmetry.

## Shared-path requirement

This changes the central Lists persistence seam, so regression-check every materially affected existing consumer:

- Lists overview/detail UI;
- add/create/remove/check/edit list skills;
- meal-plan/shopping-list projection writes;
- manual reorder behaviour;
- archive/restore behaviour where it touches the same entities;
- #908 plain-text share/copy;
- #1489 widget/update path.

Do not audit unrelated app persistence.

## Behaviour requirements

- existing local-only lists continue to work exactly as before from the user's perspective;
- new stable IDs are generated locally without network access;
[...]

Issue #928, opened 2026-05-18
Lists: two-level hierarchical parent/child items


NickMonrad:
Parent epic: #1141 — Lists UX and structured list management
Sync roadmap dependency: #1490
Depends on: #1491, #1492
Blocks feature-complete sync validation in: #1493, #1494, #1539

## Purpose

Add the existing planned **two-level parent/child hierarchy** to Jandal Lists, using the stable identity, effective-hierarchy normalisation and sync-safe ordering model defined by #1491 and implemented by #1492.

This issue owns Jandal hierarchy behaviour and UX. It does not implement transport, encrypted package exchange or CalDAV/Nextcloud code.

## Delivery precondition and source of truth

Implementation starts from `main` only after #1492 is merged.

Authoritative inputs:

- `docs/specs/shared-lists-sync.md` — identity, placement, ordering and effective-hierarchy rules;
- `docs/UX_PATTERNS.md` — reuse existing interaction/visual patterns;
- the #1492 `ListMutationRepository`, `parentItemId`, `orderKey`, placement stamp and `EffectiveHierarchyNormalizer` implementation.

Do not introduce a second hierarchy field, local-row parent identity, alternate ordering model or replacement drag library. Extend the existing `sh.calvin.reorderable:2.4.3` list-item interaction.

No additional Room hierarchy schema is expected: #1492 already supplies the required persistent placement state. Add schema only if the merged implementation proves it is genuinely necessary.

### Hierarchy depth and effective state

Support exactly two levels:

- top-level parent/standalone item;
- direct child.

Do not support arbitrary/deeper nesting.

Parent relationships use the globally stable `itemId`, never a device-local/autoincrement Room row ID.

Render and operate on the **effective hierarchy** from the #1491/#1492 normalisation rules, not raw `parentItemId` alone. A remotely received requested placement that is suppressed by normalisation is presented as top-level; local UI actions must not deliberately create invalid requested placements.

A parent may have multiple direct children.

### Visual presentation

- Children are visibly indented beneath their effective parent while preserving the existing item typography and primary row actions.
- Parent and child rows remain in one continuous group; do not add expandable/collapsible tree state.
- Child indentation must not collapse the item text into an impractically narrow column. Keep the content area readable on phone-width layouts.
[...]

PR #1540, opened 2026-09-15
docs(sync): define shared Lists sync contract

### LINKED ISSUE #1491: Shared sync: define transport-independent collection and change contract
Parent: #1490 — local-first shared Lists sync foundation
Parent Lists epic: #1141
Related hierarchy implementation: #928

## Purpose

Define the **smallest public, transport-independent data contract** needed for Jandal to synchronise a shared list correctly across devices.

This is a design/specification issue. Do not implement Room migration, network transport, backend services, billing, accounts, deployment or product packaging here.

Lists are the first concrete consumer. Do not generalise beyond what is required to make Lists safe for later cross-device synchronisation and the existing two-level hierarchy planned in #928.

### Stable identity

Specify locally generated globally stable identifiers for:

- collection/list identity;
- item identity;
- change/operation identity;
- actor/device identity used only for deterministic ordering/attribution.

Use UUIDv4-style opaque strings unless implementation evidence requires an equivalent representation. Existing Room `Long` primary keys remain local implementation details and must never become cross-device identity.

Display names are never identity.

### Version/conflict stamps

Use a persisted logical/Lamport-style version stamp such as `(counter, actorId)` for synchronised field changes.

- Local changes increment the actor's persisted counter.
- Receiving a remote stamp advances the local counter beyond the observed value.
- Same-field conflicts choose the greater deterministic stamp, with actor ID as the tie-breaker.
- Existing wall-clock `updatedAt` values are not conflict authority.

Prefer **field-level merge semantics** over whole-record last-writer-wins so independent changes such as editing item text on one device and completing it on another can both survive.

Do not introduce a generic CRDT framework unless concrete list behaviour cannot converge correctly without one.

### Change model

Define the minimum explicit mutations required for current Lists behaviour and #928 compatibility:

- create list;
- rename list;
- delete list;
- restore list;
- create item;
- edit item text;
- set/unset due date;
- check/uncheck item;
- delete item;
- restore item;
- reorder item;
- parent/unparent/reparent item using stable item identity.

The same `changeId` delivered more than once must be idempotent.

Hierarchy-driven cascades such as completing a parent are represented as the resulting explicit item changes rather than introducing a second hierarchy-specific sync protocol.

### Synchronised versus local-only fields

The contract must distinguish shared collection state from device-local presentation/automation state.

Synchronised list state includes at minimum:

- canonical shared title;
- stable collection identity;
- deletion/restoration state;
- immutable creation metadata needed for deterministic presentation where applicable.

Synchronised item state includes at minimum:

- stable item identity;
- text;
- checked/completed state;
- due date;
- parent item identity;
- sync-safe sibling order;
- deletion/restoration state;
- immutable creation metadata needed for normal sort behaviour.

Current device-local fields remain local unless a later issue explicitly changes the product behaviour:

- list pinned state;
- list archived state;
- local list-overview display order;
- item favourite state;
- notification/alarm time;
- local display-name alias used only to disambiguate otherwise-identical collection titles.

When an imported/synchronised record has no local value, use neutral defaults rather than copying another device's local state:

- imported list: unpinned, unarchived, appended to the local overview order;
- imported item: not favourite, no local notification/alarm;
- no other local-only behaviour is inferred from the sender.

### Canonical title and local display alias

A synchronised collection's canonical title is separate from its optional local-only display alias.

If an incoming collection title collides with an unrelated local list name:

- never merge by name;
- never silently rename the remote/shared collection;
- automatically allocate a unique local display alias, using the source/provider label when available (for example `Shopping (Nextcloud)`, then `Shopping (Nextcloud 2)` if required);
- keep the canonical shared title unchanged;
- avoid a blocking conflict-resolution dialog;
- surface at most a lightweight one-time notice with an optional Rename action.

An explicit user rename of the bound/shared list is a real synchronised rename and should clear/replace the generated local alias according to the implementation contract.

### Deletion, tombstones and restore

Use tombstone semantics rather than immediately losing synchronised identity.

- A deletion stamp wins over stale edits/checks/reorders.
- A newer explicit restore operation may reactivate the same stable identity.
- Stale offline data must not resurrect a tombstone accidentally.
- Do not automatically garbage-collect tombstones in the first implementation; safe compaction requires evidence that old peers cannot replay unseen stale state.

A user-facing Jandal Trash/Recently Deleted screen is **not required by this issue**. The sync format must simply support restore so a later local UX or a provider-side restore can reactivate an item without changing identity.

### Ordering

Define a sync-safe sibling order representation that does not depend on device-local integer positions.

Preferred first design: a canonical arbitrary-precision fractional order key:

- append after `5` → `6`;
- insert between `5` and `6` → `5.5`;
- insert between `5` and `5.5` → `5.25`;
- ties resolve deterministically using stable item identity.

Parent and top-level sibling groups use the same rule independently.

The authoritative spec may choose an equivalent representation if it proves simpler while preserving concurrent insert/reorder convergence without routine whole-list renumbering.

### #928 hierarchy compatibility

The contract must support the existing #928 two-level model without adding deeper nesting:

- `parentItemId` refers to a stable item identity, not a Room row ID;
- parent relationship is independently versioned from order;
- only parent → child depth is valid;
- deleting/capturing a parent does not delete its children; required unparent/reparent mutations are explicit;
- a child received before its referenced parent must remain visible (temporarily top-level) rather than being hidden or dropped.

#928 remains authoritative for hierarchy UX and completion/grouping behaviour. #1491 defines only the sync-safe representation required so #928 does not cause a later destructive schema redesign.

### Snapshot and incremental change boundary

Treat Room current state as authoritative. Do **not** turn Lists into an event-sourced database.

- First exchange/import can use a full collection snapshot including required tombstones.
- Incremental exchange can use recorded changes plus a per-actor contiguous checkpoint.
- If a sender cannot satisfy a requested checkpoint because older changes have been compacted, fall back to a current full snapshot.

This keeps the local change log bounded while retaining deterministic reconciliation.

### Sharing/package boundary

Define a versioned encrypted package representation sufficient for #1493 without defining a hosted-service API.

The contract should specify a `SharedCollectionPackageV1`-style envelope with:

- format/package identity;
- collection identity;
- key/reference identifier;
- authenticated-encryption nonce/metadata;
- encrypted collection snapshot and/or changes;
- tombstones and checkpoint state;
- minimum membership/invite metadata required by the client.

List/item content must not be exposed in the plaintext envelope. Prefer a standard authenticated-encryption primitive such as AES-256-GCM using a random collection key.

Key-server, hosted account, key rotation/revocation service and commercial packaging design remain out of scope.

### Current Room mapping requirement

The design must map onto Jandal's existing Room-backed Lists implementation rather than inventing a parallel store.

Current sync-relevant writes originate from multiple places, including `ListsViewModel`, `NativeIntentHandler`, and `MealPlanSessionRepository`. #1492 therefore must ensure each synchronised state mutation and its change/outbox record commit through **one atomic Room transaction seam**, so UI-, voice- and meal-plan-created changes cannot bypass synchronisation.

Read-only DAO consumers do not need to move behind a new abstraction solely for consistency.

## Deliverable

Add one focused authoritative subsystem spec under the existing `docs/specs/` structure, and register it in `docs/SPEC_INDEX.md`.

The spec must describe:

- object/stable-ID model;
- field/version model;
- change model;
- conflict rules;
- synchronised versus local-only state and defaults;
- canonical title/local display-alias rules;
- hierarchy representation;
- ordering model;
- deletion/tombstone/restore rules;
- snapshot/checkpoint/change-log boundary;
- encryption/package boundary;
- mapping to the current Lists schema and mutation paths;
- explicit non-goals.

The document becomes the source of truth for #1492, #928, #1493, #1494 and provider integrations such as #1539.

## Acceptance criteria

- [ ] Globally stable list, item, change and actor IDs are specified without a central allocator.
- [ ] Synchronised fields use deterministic logical/version stamps rather than wall-clock `updatedAt` as conflict authority.
- [ ] Current list mutations, restore operations and #928 parent/unparent operations have explicit change representations.
- [ ] Duplicate change delivery is idempotent.
- [ ] Local-only fields and neutral import defaults are explicitly defined.
- [ ] Canonical shared title is distinct from a local-only display alias, and name collisions never cause identity matching or forced remote rename.
- [ ] Tombstone/delete and explicit restore behaviour are specified; stale changes cannot accidentally resurrect deleted data.
- [ ] Sync-safe ordering is specified and compatible with #928 two-level hierarchy.
- [ ] Deterministic field-level MVP conflict rules are documented.
- [ ] Snapshot/checkpoint semantics allow a bounded local change log without requiring permanent event history.
- [ ] A versioned encrypted shared-package boundary is defined without a hosted-service API.
- [ ] The design maps cleanly onto the current Room-backed Lists model and identifies the required atomic mutation/change-recording seam for #1492.
- [ ] No generic CRDT/distributed-data framework is introduced without demonstrated need.
- [ ] No backend, subscription, billing, deployment or commercial strategy appears in the design.

## Out of scope

- Room migration implementation;
- hierarchy UX implementation (#928);
- user-facing Trash/Recently Deleted UI;
- network/nearby transport;
- production internet sync;
- account/social graph design;
- notes, memory, chat, profile or attachment sync;
- billing/entitlements;
- backend/service implementation.

NickMonrad:
## Summary

Adds the authoritative transport-independent Shared Lists sync contract required by #1491 and registers it in `docs/SPEC_INDEX.md`.

The spec defines the minimum public contract needed by the active Lists roadmap:

- stable collection/item/change/actor identity;
- actor-global Lamport conflict clock plus contiguous per-actor, per-collection delivery sequences/checkpoints;
- field-level deterministic merge semantics;
- sticky tombstones plus explicit restore;
- synchronised versus device-local fields and neutral import defaults;
- canonical shared title versus generated local display alias for non-blocking name collisions;
- two-level hierarchy representation compatible with #928;
- atomic `(parentItemId, orderKey)` placement semantics plus deterministic cross-item hierarchy normalisation for cycles/depth conflicts;
- full-snapshot/incremental-checkpoint boundary with a bounded change log;
- versioned AES-256-GCM encrypted package/invite boundary;
- mapping onto the current Room entities and all known sync-relevant writers;
- the required atomic Room mutation + outbox seam for #1492.

## Scope

Docs/specification only. No Room migration, Kotlin implementation, hierarchy UX, transport, Nextcloud code, hosted service or commercial strategy is included.

## Repository grounding

The contract is explicitly mapped to the current `ListNameEntity`, `ListItemEntity`, `ListsViewModel`, `NativeIntentHandler`, and `MealPlanSessionRepository` write paths so #1492 cannot make only one mutation surface sync-aware.

## Review remediation

Addresses review #5204593657:

- delivery sequencing/checkpoints are now scoped to `(actorId, collectionId)`, while the Lamport conflict clock remains actor-global;
- cross-item placement conflicts derive an acyclic two-level effective hierarchy using a deterministic ordering over placement stamps and stable IDs;
- #1494 validation now explicitly covers opposite-direction cycles, depth >2 conflicts, delivery-order independence, and interleaved changes across multiple collections.

## Self-review

- no TODO/TBD placeholders;
- no hosted/commercial/private-strategy content;
- hierarchy remains exactly two levels;
- duplicate text/name is never identity;
- delete cannot be undone by stale ordinary edits;
- parent and sibling order are stamped as one requested placement field;
- invalid cross-item hierarchy combinations are normalised deterministically rather than rejected by arrival order;
- local-only fields are excluded from package/sync state;
- spec is indexed as a canonical subsystem contract.

Closes #1491


NickMonrad (commented):
**Review result: changes required.** Two material contract defects need correction before this spec becomes authoritative.

1. **High — `docs/specs/shared-lists-sync.md`, “Per-actor source sequence” (~118–135) and “Checkpoint / Incremental exchange” (~548–573): collection-scoped packages cannot use one actor-global contiguous sequence.** The spec says every locally originated change increments one per-actor `sourceSequence`, while exchange/checkpoints are for one `collectionId`. If actor A changes collection X (seq 1), collection Y (seq 2), then X again (seq 3), an X-only package can deliver 1 and 3 but can never satisfy the required contiguous checkpoint through 2 because seq 2 belongs to another collection. Incremental sync for X therefore stalls or repeatedly falls back to snapshots. This violates the bounded incremental checkpoint acceptance criterion. **Smallest remediation:** scope the delivery sequence/checkpoint to the collection, e.g. contiguous sequence per `(actorId, collectionId)` (while keeping the Lamport clock actor-global), and update the change/checkpoint wording accordingly.

2. **High — `docs/specs/shared-lists-sync.md`, “Two-level hierarchy contract / Valid relationships” (~440–454) plus remote reconciliation/failure rules (~758–778): concurrent cross-item placements have no deterministic invariant-repair rule.** Atomic placement only resolves conflicts on the *same item*. Two devices can independently make A a child of B and B a child of A; both writes are locally valid and target different placement fields, so normal field reconciliation retains both and creates a cycle. Likewise, concurrent `P -> Q` and `C -> P` can create depth 3. The current instruction to enforce #928 invariants / reject hierarchy deeper than two levels is apply-order dependent and does not say which placement wins, so peers can reject/derive different outcomes and fail #1494 hierarchy convergence. **Smallest remediation:** specify one deterministic cross-item hierarchy repair rule using existing placement stamps/stable IDs (no generic CRDT), and add cycle/depth-conflict cases to the downstream convergence validation requirements.

The docs-only CI run is currently red because the `Set up Android SDK` step failed before build/lint/tests; Docs Drift Check passed. That appears unrelated to these document changes, but the two contract issues above are merge blockers.

NickMonrad:
Remediation for review #5204593657 is now on the PR.

1. Delivery sequencing/checkpoints are collection-scoped: the Lamport conflict clock remains actor-global, while `sourceSequence` is contiguous per `(actorId, collectionId)`. The spec includes the X/Y/X example and collection-scoped snapshot/checkpoint rules.
2. Cross-item hierarchy conflicts now use deterministic effective-hierarchy normalisation: candidate parent edges are ordered by placement stamp, then stable IDs, and accepted only when they preserve an acyclic two-level forest. Requested placements remain synchronised state; normalisation is derived, so delivery order does not decide the result.
3. #1494 validation now explicitly covers interleaved multi-collection sequencing, A↔B cycles, depth >2 conflicts, and applying the same placement set in different delivery orders.

Focused diff from the reviewed head (`e0e2019`) to the current head changes only `docs/specs/shared-lists-sync.md` (85 additions / 46 deletions). Ready for remediation-only re-review.

NickMonrad (commented):
Re-review of the remediation for review #5204593657 only.

Both previously reported contract defects are resolved:

1. Delivery sequencing/checkpoints are now scoped to `(actorId, collectionId)` while the Lamport conflict clock remains actor-global, so interleaved changes across different lists cannot create false checkpoint gaps.
2. Cross-item hierarchy conflicts now derive one deterministic acyclic two-level effective hierarchy from reconciled requested placements. The fixed candidate ordering and suppression rules remove arrival-order dependence for opposite-direction cycles and depth >2 conflicts, while retaining the underlying requested placement registers for later reconciliation.

I found no material defect introduced by these remediation changes. The added downstream validation cases cover the original failure scenarios.

Docs Drift Check passes on the remediation head. The general CI workflow still fails during Android SDK setup before build/lint/unit tests, matching the unrelated infrastructure failure seen before this remediation.

From the re-review scope, PR #1540 is ready to merge.

Issue #883, opened 2026-05-16
feat: reduce repeated meals across meal plans


lokhor:
Parent epic: #1139 — Meal planning experience and deterministic recipe workflow

## Summary

Reduce repeated or near-identical meals across different meal plans by making variety the default behaviour.

During testing, the meal planner often falls back to the same small cluster of meals across separate plans. Even when the exact title changes slightly, the result can still feel like the same recipe shape repeated too often. The planner should prefer variety by default instead of repeatedly surfacing the same chicken skillet / beef stir-fry style meals.

## Problem

Today the deterministic planner can generate multiple plans that feel too similar across time:
- exact repeats of prior meals
- near-duplicates with slightly different wording
- repeated protein + cooking-shape combinations across adjacent plans

This makes the planner feel stale and reduces trust that it is genuinely planning around the user's preferences instead of reusing a narrow default set.

## Goal

Make the meal planner prefer reasonable novelty by default while preserving the existing deterministic, app-owned architecture.

### 1. Track recent meal history
Use canonical persisted planner data to derive a bounded recent-history set, for example:
- recent meal titles
- normalized meal identities
- recent protein + cooking-shape combinations
- meals used in the last N plans or last N days

### 2. Bias plan generation away from recent repeats
When generating a new high-level plan:
- include a compact avoid-list / recent-history summary in the bounded generation prompt
- tell the model to prefer variety and avoid recently used meals unless the user explicitly wants familiar meals

### 3. Add app-side repeat detection
Do not rely only on prompt wording.

Add deterministic checks for:
- exact title matches
- normalized title matches
- obvious near-duplicates such as the same protein + same meal pattern

If a plan comes back too repetitive, regenerate the overlapping days before presenting the plan to the user.

### 4. Leave room for user-facing planning modes later
This issue does not need full preference controls yet, but it should leave a clean path for later options such as:
- prefer variety
- balanced
- prefer familiar meals

## Acceptance criteria
- [ ] New meal plans soft-avoid exact repeats from recent meal plans by default
- [ ] The planner also reduces obvious near-duplicates, not just exact title collisions
- [ ] Repeat-avoidance is derived from canonical persisted planner data, not chat-history reconstruction
- [ ] The implementation remains bounded and deterministic in the app-owned planner flow
- [ ] The change leaves a clean path for future user-facing planning modes such as `prefer variety` or `prefer familiar meals`

## Related
- Deterministic planner foundation: #859
[...]

PR #1547, opened 2026-09-20
feat(#1539): sync Lists with Nextcloud Tasks via CalDAV

### LINKED ISSUE #1539: Lists: sync selected lists with Nextcloud Tasks via CalDAV/VTODO
Parent: #1490 — local-first shared Lists sync foundation
Depends on: #1491, #1492, #928, #1494
Related: #1141, #1489, historical #315, #314

## Purpose

Add a **self-managed Nextcloud sync transport for Jandal Lists** using the standard CalDAV `VTODO` model used by Nextcloud Tasks.

This is a concrete post-foundation integration, not a second list store or a separate conflict model. Jandal remains local-first: Lists UI, list skills and widgets continue reading the existing Room-backed data, while a Nextcloud adapter exchanges selected list changes with a user-configured Nextcloud server.

The target is feature-complete interoperability for Jandal's implemented Lists model, including the #928 two-level parent/child hierarchy.

The motivating use case is straightforward: a user who already runs Nextcloud should be able to use the same shopping/todo lists from Jandal, the Nextcloud Tasks web UI and compatible Nextcloud Tasks Android clients without relying on Google services.

## Validated integration target

The Android app `com.nextcloud.tasks` documents that it synchronises against Nextcloud using CalDAV/VTODO and supports offline task lists. The Nextcloud Tasks server app supports subtasks and interoperable CalDAV clients.

For hierarchy, iCalendar `RELATED-TO` is the standard relationship property for `VTODO`. A child VTODO can reference the parent's `UID` using the default/explicit `PARENT` relationship.

References:

- Nextcloud Tasks Android: https://play.google.com/store/apps/details?id=com.nextcloud.tasks
- Android client source: https://github.com/SebiShepherd/nextcloud_tasks_android
- Nextcloud Tasks server app: https://github.com/nextcloud/tasks
- iCalendar `RELATED-TO`: https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.4.5
- Nextcloud CalDAV endpoint guidance: https://docs.nextcloud.com/server/stable/admin_manual/issues/general_troubleshooting.html#service-discovery

Jandal must integrate through CalDAV. Do **not** read or write Nextcloud database tables or depend on the Android app's local database.

## Required prerequisite state

Implement this only after the transport-independent Lists work has proven the required local semantics:

- #1491 defines stable identity, field/version merge, deletion/restore, ordering, canonical title/local alias and hierarchy representation;
- #1492 provides sync-ready Room persistence/change recording and the authoritative atomic mutation seam;
- #928 implements Jandal's two-level parent/child UX and behaviour on that model;
- #1494 proves flat-list and hierarchy convergence/idempotency through the normal local persistence seam.

Do not bypass those seams with a Nextcloud-specific parallel model.

### 1. One Nextcloud account

Add the smallest configuration needed for one self-managed Nextcloud account:

- server/base URL;
- username;
- app password/password credential suitable for CalDAV;
- connection test with a truthful success/failure state.

Store the credential using Jandal's existing secure secret-storage pattern/Android-backed encrypted storage. Never put credentials in Room list entities, logs, diagnostics, issue evidence or exported sync data.

Do not build a general account framework or multi-account UI in this issue.

### 2. CalDAV discovery

Use standard CalDAV discovery rather than hard-coding a user's server layout.

Support normal Nextcloud discovery via `/.well-known/caldav` and/or the documented `/remote.php/dav` flow, then discover the user's calendar home and collections that support `VTODO`.

A failed or partially configured server must produce an actionable error rather than silently falling back to local-only behaviour while claiming sync succeeded.

### 3. Explicit list binding

Only explicitly selected lists participate in Nextcloud sync.

For the first slice support these safe paths:

- **Import a Nextcloud task list** as a new Jandal list and bind it one-to-one to that remote CalDAV collection.
- **Publish a Jandal list** by creating a new remote task collection and binding it one-to-one.

Do not implement arbitrary merge of an already-populated local list with an already-populated remote list in the first slice.

Persist the provider binding independently of editable list names.

### 4. Canonical title, local alias and local-only defaults

The remote/shared collection title remains canonical sync state. Jandal's local display alias is device-local only.

If an imported Nextcloud task-list title collides with an unrelated local Jandal list:

- do not merge by name;
- do not rename the Nextcloud collection merely to satisfy the local collision;
- automatically use a unique local display alias such as `Shopping (Nextcloud)`, then `Shopping (Nextcloud 2)` if required;
- avoid a blocking conflict dialog;
- show at most a lightweight one-time notice with an optional Rename action.

Apply #1491 neutral defaults to imported local-only state:

- list starts unpinned/unarchived and is appended to local overview order;
- item favourite is false;
- item local notification/alarm is unset.

Do not push those device-local fields to Nextcloud.

### 5. VTODO mapping

Map the Jandal fields required for feature-complete current Lists behaviour:

- item text ↔ `SUMMARY`;
- active/completed state ↔ VTODO status/completion fields;
- due date ↔ VTODO due date;
- Jandal parent/child relationship ↔ `RELATED-TO` referencing the parent VTODO `UID` with parent semantics;
- deletion ↔ remote resource delete plus Jandal tombstone semantics;
- provider-side restoration/reappearance of a known bound resource ↔ an explicit Jandal restore where stable provider identity permits it;
- remote `UID`, resource `href` and `ETag` ↔ provider-binding metadata required for stable remote identity and optimistic concurrency.

Jandal's own stable list/item IDs remain authoritative local sync identities. Do not replace them with display text, Room row IDs, remote URLs or remote UIDs.

When reading an existing remote task, create/persist a provider binding to its VTODO identity rather than treating matching text as identity.

Jandal enforces its #928 two-level hierarchy even if another CalDAV client can represent deeper nesting. Do not expand Jandal to arbitrary depth in this issue.

### 6. Preserve unsupported remote data

A Jandal edit to a known field must not unnecessarily destroy unrelated VTODO properties produced by Nextcloud Tasks or another CalDAV client.

In particular:

- preserve unrelated properties where the chosen library permits safe round-tripping;
- preserve relationship metadata that Jandal does not actively modify;
- do not flatten a supported #928 parent/child relationship during ordinary edits.

If remote data expresses hierarchy deeper than Jandal supports, do not silently rewrite/drop the unknown relationship merely because Jandal only renders two levels. Keep #928 as the local behaviour limit and preserve unsupported remote metadata where safe.

### 7. Local-first sync behaviour

The bound list must remain fully usable with the server unavailable.

- Local operations commit to Room first through #1492's authoritative mutation seam.
- Existing #1492 change/outbox state records pending work.
- The Nextcloud adapter sends pending changes when connectivity is available.
- Remote changes are translated into normal #1491 changes and applied through the same authoritative local persistence/update seam.
- Lists UI and skills never read directly from CalDAV.
- #1489 widgets remain local-data-only and refresh after Room changes normally.

Do not block list CRUD/hierarchy operations on a network round trip.

### 8. Sync triggers

Use the smallest practical Android scheduling model:

- explicit/manual refresh for deterministic testing and recovery;
- refresh when the bound list/app becomes active where appropriate;
- bounded WorkManager background sync for eventual convergence;
- enqueue/trigger sync after local mutations without creating an always-on service.

Do not add push infrastructure or a long-running network foreground service.

### 9. Concurrency and conflicts

Use CalDAV `ETag`/conditional updates so Jandal does not blindly overwrite a remote resource that changed since it was read.

When the server reports a concurrent modification:

1. fetch the current remote object;
2. translate it into the normal Jandal change representation;
3. reconcile using the deterministic field/version rules from #1491;
4. persist the converged state locally;
5. retry the minimum required remote update if needed.

Do not invent a second set of Nextcloud-specific conflict rules.

## Required interoperability validation

Use a real self-managed Nextcloud test instance, but do **not** publish its hostname, username, credentials, app password or other private deployment details in GitHub evidence.

Validate against both the Nextcloud Tasks server/web UI and the installed Android `com.nextcloud.tasks` client where practical.

Required scenarios:

1. Create/publish a Jandal list; it appears as a task list in Nextcloud.
2. Add an item in Jandal; it appears in Nextcloud Tasks.
3. Complete/uncomplete the task in Nextcloud Tasks; Jandal converges to the same state.
4. Rename/edit task text from each side and verify propagation.
5. Delete from each side and verify the Jandal tombstone/remote deletion path.
6. Restore a previously deleted known task from the provider side where the tested Nextcloud version exposes restore, and verify Jandal restores the same logical item without duplication when provider identity permits it.
7. Create a parent with child items in Jandal; verify Nextcloud Tasks represents them as subtasks and preserves their relationship.
8. Create/move a subtask in Nextcloud Tasks within Jandal's supported two-level model; verify Jandal converges to the same parent/child relationship.
9. Reorder within a parent group and verify deterministic Jandal ordering after round-trip, without requiring unsupported provider ordering semantics to be invented.
10. Make a Jandal change offline, reconnect and verify it is eventually pushed.
11. Make a remote change while Jandal is offline, reconnect and verify it is pulled.
12. Exercise one genuine concurrent-edit conflict and prove deterministic convergence rather than blind overwrite.
13. Re-run sync with no changes and prove it is idempotent/no-op.
14. Import a remote list whose title collides with an unrelated local list and verify a non-blocking local alias such as `Shopping (Nextcloud)` is used without renaming the remote collection.
15. Restart Jandal and prove account/list bindings, hierarchy bindings and pending sync state survive correctly.

Capture app commit/build, Nextcloud server version, Nextcloud Tasks app version and normalised final list state. Do not capture private server identifiers or credentials.

## Tests

Add focused automated coverage for:

- CalDAV discovery parsing;
- VTODO parse/serialise round trips for supported fields;
- `RELATED-TO` parent mapping using VTODO `UID`;
- provider binding persistence;
- import-new-list and publish-new-list paths;
- local title-collision alias allocation;
- local-only neutral defaults;
- local create/edit/complete/delete/restore/hierarchy changes queued while offline;
- remote create/edit/complete/delete/restore/hierarchy application;
- duplicate pull/push idempotency;
- ETag conflict/refetch/reconcile path;
- unknown/unmapped VTODO property preservation where applicable;
- credential redaction from logs/errors;
- list UI/skills continuing to consume Room rather than transport state.

Prefer protocol fixtures/local fake HTTP responses for automated tests. The private real server is for bounded integration/device validation, not a CI dependency.

## Acceptance criteria

- [ ] One user-configured Nextcloud account can be connected securely for CalDAV task sync.
- [ ] Jandal discovers VTODO-capable Nextcloud task collections without direct database access.
- [ ] A remote task list can be imported as a new bound Jandal list.
- [ ] A local Jandal list can be published as a new bound Nextcloud task list.
- [ ] Only explicitly bound lists synchronise.
- [ ] Item create, text edit, complete/uncomplete, delete and supported restore synchronise bidirectionally.
- [ ] Due date synchronises.
- [ ] #928 two-level parent/child relationships round-trip via standard VTODO relationship metadata without flattening.
- [ ] Existing remote task identity is tracked by provider metadata rather than item text.
- [ ] Jandal's stable local sync IDs remain independent of remote URLs/display names/UIDs.
- [ ] Imported local-only fields use #1491 neutral defaults and are not pushed to Nextcloud.
- [ ] Local title collisions use a unique local display alias without name-based merge, remote rename or blocking conflict dialog.
- [ ] Local list and hierarchy operations remain fully usable offline.
- [ ] Pending local changes synchronise after connectivity returns.
- [ ] Remote changes enter through the normal Room-backed persistence/update seam.
- [ ] CalDAV ETags/conditional requests prevent blind lost-update overwrites.
- [ ] Concurrent modification is reconciled using #1491 rules.
- [ ] Repeated sync with unchanged state is idempotent.
- [ ] Existing Lists UI, skills and #1489 widget path require no transport-specific data access.
- [ ] Unmapped VTODO data is not destructively discarded by ordinary supported-field edits where safe preservation is possible.
- [ ] Credentials and private server details do not appear in logs, diagnostics or public test evidence.
- [ ] Real-device interoperability is proven against the configured test Nextcloud instance and Nextcloud Tasks client/web UI.

## Out of scope

- hosted Jandal relay/service implementation;
- billing, subscriptions, pricing or entitlements;
- Nextcloud server deployment/administration;
- direct Nextcloud database access;
- Nextcloud Notes sync;
- Nextcloud Deck integration;
- CardDAV/contact sync;
- `VEVENT` calendar-event sync;
- multiple Nextcloud accounts in the first slice;
- arbitrary merge UX for two already-populated local/remote lists;
- hierarchy deeper than #928's two levels;
- rich VTODO descriptions/categories/attachments/reminders unless separately required;
- user-facing Jandal Trash/Recently Deleted UI;
- general CalDAV framework abstractions for hypothetical providers;
- making this a v0.1 launch blocker.

## Priority

**Active high-priority Lists workstream.** The v0.1 launch work may remain paused while this sequence is delivered, but this issue is still not itself a v0.1 release blocker.

lokhor:
Closes #1539

## What changed
- Adds one-account Nextcloud configuration with Android Keystore-backed encrypted app-password storage.
- Discovers CalDAV principals and VTODO-compatible calendars using standard `.well-known/caldav`, PROPFIND, REPORT, MKCALENDAR, PUT, and DELETE flows.
- Imports remote task lists idempotently and publishes explicitly selected local Lists.
- Stores stable collection/item bindings and provider metadata in Room; local Room data remains authoritative.
- Maps VTODO SUMMARY, STATUS, COMPLETED, DUE, RELATED-TO, and `X-JANDAL-ORDER` while preserving unknown VTODO properties and unrelated relationship lines.
- Uses the existing List outbox plus manual, active-app, and WorkManager sync triggers.
- Uses conditional ETags with pull/retry conflict handling, remote deletion tombstones, and restore-on-reappearance behavior.
- Adds Settings UI, migration coverage, protocol/parser tests, and the implementation spec.

## Verification
- `./gradlew :core:memory:testDebugUnitTest --tests 'com.kernel.ai.core.memory.nextcloud.*'`
- `./gradlew testDebugUnitTest`
- `./gradlew :core:memory:compileDebugKotlin :feature:settings:compileDebugKotlin :app:compileDebugKotlin`
- `./gradlew assembleDebug`
- `./gradlew lint`
- `./gradlew :core:memory:lintDebug :feature:settings:lintDebug :app:lintDebug`
- `./gradlew :core:memory:connectedDebugAndroidTest` (both connected devices; migration suite passed)
- Installed debug APK and manually verified Settings → Nextcloud Tasks plus empty-field validation on a connected Samsung device.

## Remaining environment gate
No configured real Nextcloud instance or app-password is available in this development environment, so live interoperability evidence could not be collected without requesting credentials. Before merge, exercise import, publish, unknown-property preservation, local/remote deletion and restoration, and an ETag conflict against a real self-managed Nextcloud Tasks calendar.

No hosted relay/backend or production internet sync service is introduced.


NickMonrad (changes requested):
One material defect remains in the remediation.

**MEDIUM — `syncAll()` now swallows coroutine cancellation and continues syncing later bindings**

**Location:** `core/memory/src/main/java/com/kernel/ai/core/memory/nextcloud/NextcloudSyncAdapter.kt`, `syncAll()`, the new `catch (error: Exception)` around each `syncBoundCollection(...)` call.

**Failure scenario:** WorkManager or the caller cancels a sync while one binding is in a suspend/network operation. `CancellationException` is an `Exception`, so this new catch converts it to a `NETWORK` failure and continues iterating over the remaining bindings. A cancelled job can therefore keep doing CalDAV reads/writes after its lifecycle has ended.

**Why this is a defect:** the remediation intentionally broadens the shared sync path so one failed collection does not block later bindings, but cancellation is control flow rather than a collection failure. Swallowing it violates coroutine/lifecycle correctness and can cause network/resource work after cancellation.

**Smallest remediation:** add `catch (error: CancellationException) { throw error }` before the generic `Exception` catch (and import `kotlinx.coroutines.CancellationException`), then add a focused regression test proving cancellation stops before later bindings are processed.

The intended remediation for quoted/bare ETags, authenticated write-403 permission mapping, read-only pull-with-pending preservation, and continuation to later bindings otherwise looks correct. CI at this head is green.

lokhor:
Cancellation remediation pushed in commit f5fc8468.

- Added kotlinx.coroutines.CancellationException handling before generic Exception in NextcloudSyncAdapter.syncAll().
- Cancellation is rethrown immediately; later bindings are not processed.
- Preserved cancellation through the sync result guard so it cannot become NETWORK failure.
- Added regression coverage for propagation and later-binding suppression.

Validation passed:
- ./gradlew :core:memory:testDebugUnitTest --tests 'com.kernel.ai.core.memory.nextcloud.NextcloudSyncAdapterTest'\n- ./gradlew :core:memory:testDebugUnitTest\n- ./gradlew :app:testDebugUnitTest\n- ./gradlew :app:assembleDebug\n- git diff --check\n\nNo physical Nextcloud acceptance rerun; this change is cancellation-only. Do not merge — wait for review.

lokhor:
CI for f5fc84680338677f301c536dc8377ec4974960d6 passed:
- Build & Test — success
- Docs Drift Check — success

NickMonrad (approved):
Re-review of the remediation only.

The previous cancellation finding is resolved at `f5fc84680338677f301c536dc8377ec4974960d6`:
- `syncAll()` explicitly rethrows `CancellationException` before its ordinary failure handlers.
- `guardedResult()` also preserves cancellation instead of converting it to `NETWORK`.
- The focused regression test proves cancellation propagates and later bindings are not processed.

The remediation is narrowly scoped and introduces no new material defect. CI #3217 and Docs Drift Check #883 are green.

Ready to merge from the reviewed scope.

